PR #21, opened 2024-03-04
Add dependencies to dune-project file

panglesd:
Removed some vendoring in order not to increase the dependencies.


dune-project
```diff
@@ -11,15 +11,15 @@
 
 (maintainers "Paul-Elliot <peada@free.fr>" "Jules Aguillon <jules@j3s.fr>")
 
-(license LICENSE)
+(license MIT)
 
 (documentation https://url/to/documentation)
 
 (package
```
Julow, line 18: At this point, I think we should correctly fill the dune project. I don't think external contributors would be confident in doing that for us.

vendor/lwd/dune-project
```diff
@@ -24,19 +24,6 @@
           (cbor :with-test) ; for the examples
           (containers :with-test))) ; for the examples
 
-(package
- (name tyxml-lwd)
```
Julow, line 28: Removing the unnecessary parts will interfere with git subtree and can be avoided with `(vendored_dirs vendor)`: https://dune.readthedocs.io/en/stable/dune-files.html#vendored-dirs

PR #29, opened 2024-03-06
A more representative example file

### LINKED ISSUE #28: A more representative example file
The file `diff.example` is nice for easily and quickly trying out the program but is not enough for testing it.

We need a new diff file that contains:

- All the different operations (for example, to test https://github.com/panglesd/diffcessible/pull/27)
- A hunk with only deleted lines
- A hunk with only added lines
- A hunk with both.

Ideally, this file would be generated by first crafting two directories with slightly different files in them and running on them `git diff --no-index`.

Dezzy12:
Made some changes to the `example.diff` file.
Fixes #28 

@panglesd @Julow


example.diff
```diff
@@ -1766,3 +1766,78 @@ index 0000000..3b7560e
 +
 +let () =
 +  Alcotest.run "Patch tests" tests
```
Julow, line 1768: The existing patch is very long and a lot less interesting than the new one. Perhaps we could keep only the first 4 hunks. The other ones are very large and are about vendored libraries, these should definitely go.

example.diff
```diff
@@ -1766,3 +1766,78 @@ index 0000000..3b7560e
...
+index dfd0d1e..ce59064 100644
+--- a/dir1/file.txt
++++ b/dir2/file.txt
+@@ -1 +1 @@
+-This is the original content.
++This is the modified content.
+diff --git a/dir2/new_diff.diff b/dir2/new_diff.diff
+new file mode 100644
+index 0000000..e69de29
+
```
Julow, line 1780: Diff hunks don't normally have empty lines in between.

Julow (commented):
Looks good :) I have left two comments but this is already good quality!

Dezzy12:
@Julow
I have implemented requested changes, please review

Julow (approved):
This looks very nice! Let's merge!

Issue #17, opened 2024-02-28
Compute and display number of added and removed lines in top bar


panglesd:
The top bar currently displays the current operation shown.

It would be nice to also show the number of added and removed lines. This would have to be computed from the list of hunks.

panglesd:
For instance:

```
--- a/path/to/file
+++ b/path/to/file
```
could become:
```
--- a/path/to/file
+++ b/path/to/file
12 additions, 45 removals
```

maha-sachin:
Hi @panglesd , 
I am Mahalakshmi, part of the Outreachy Summer 2024 program, and I would like to work on this issue. I found the setup descriptions very helpful and easy to follow, here is my screengrab of `dune exec -- diffcessible  --help` 
<img width="1648" alt="Screenshot 2024-03-04 at 8 50 19 PM" src="https://github.com/panglesd/diffcessible/assets/122064139/c6d6e647-5f82-48a9-90c4-ec919ae7334a">

maha-sachin:
Thanks @panglesd, I will start working on it

maha-sachin:
Hi @panglesd & @Julow , 
I see 403 when trying to push my branch to the remote, can you help ?
```
(base) sachin@Pacharis-Laptop diffcessible % git push origin 17-display-added-removed 
remote: Permission to panglesd/diffcessible.git denied to maha-sachin.
fatal: unable to access 'https://github.com/panglesd/diffcessible.git/': The requested URL returned error: 403
```

maha-sachin:
@panglesd @Julow Ignore that comment, I found the contribution.md explaining it

Julow:
Fixed by https://github.com/panglesd/diffcessible/pull/30, thanks!

PR #30, opened 2024-03-07
Added implementation for calculating number of lines added and removed 

maha-sachin:
Added modification_counter function in interactive_viewer.ml that will provide added and removed line count, also updated formatting as per contribution document.
fix for issue: https://github.com/panglesd/diffcessible/issues/17
Sample output: 
Screenshot 2024-03-06 at 7 46 57 PM

@Julow and @panglesd

### DIFF
--- a/lib/interactive_viewer.ml
+++ b/lib/interactive_viewer.ml
@@ -60,7 +60,6 @@ let navigate z_patches (dir : direction) : unit =
   | Prev -> Lwd.set z_patches (Zipper.prev z)
   | Next -> Lwd.set z_patches (Zipper.next z)
 
-(* let pure_str s = Lwd.pure (W.string s) *)
 let quit = Lwd.var false
 
 let view (patches : Patch.t list) =
--- a/lib/interactive_viewer.ml
+++ b/lib/interactive_viewer.ml
@@ -62,41 +62,35 @@ let navigate z_patches (dir : direction) : unit =
 
 let quit = Lwd.var false
 
-let count_lines_in_hunk (addition, removal) line =
-  match line with
-  | `Their _ -> (addition + 1, removal)
-  | `Mine _ -> (addition, removal + 1)
-  | `Common _ -> (addition, removal)
-
-let calculate_additions_removals_in_hunk lines =
-  List.fold_left count_lines_in_hunk (0, 0) lines
-
-let calculate_additions_removals_in_patches patches =
-  let combine_additions_removals (add1, remove1) (add2, remove2) =
-    (add1 + add2, remove1 + remove2)
+let add_lines lines =
+  let add_line (addition, removal) line =
+    match line with
+    | `Their _ -> (addition + 1, removal)
+    | `Mine _ -> (addition, removal + 1)
+    | `Common _ -> (addition, removal)
   in
+  List.fold_left add_line (0, 0) lines
+
+let add_hunks hunks =
   List.fold_left
     (fun total_counts hunk ->
-      let additions, removals =
-        calculate_additions_removals_in_hunk hunk.Patch.lines
-      in
-      combine_additions_removals total_counts (additions, removals))
-    (0, 0) patches
+      let additions, removals = add_lines hunk.Patch.lines in
+      (fun (add1, remove1) (add2, remove2) -> (add1 + add2, remove1 + remove2))
+        total_counts (additions, removals))
+    (0, 0) hunks
 
-let calculate_addition_removal_operation z_patches : ui Lwd.t =
+let change_summary z_patches : ui Lwd.t =
   let$ z = Lwd.get z_patches in
   let p = Zipper.get_focus z in
-  let total_additions, total_removals =
-    calculate_additions_removals_in_patches p.Patch.hunks
-  in
-  let count_with_plural n singular plural =
+  let total_additions, total_removals = add_hunks p.Patch.hunks in
+  let format_plural n singular plural =
     if n = 1 then Printf.sprintf "%d %s" n singular
     else Printf.sprintf "%d %s" n plural
   in
   let operation_count =
     Printf.sprintf "%s, %s"
-      (count_with_plural total_additions "addition" "additions")
-      (count_with_plural total_removals "removal" "removals")
+      (format_plural total_additions "addition" "additions")
+      (format_plural total_removals "removal" "removals")
   in
   W.string ~attr:Notty.A.(fg lightcyan) @@ Printf.sprintf "%s\n" operation_count
 
@@ -109,7 +103,7 @@ let view (patches : Patch.t list) =
   W.vbox
     [
       operation_info z_patches;
-      calculate_addition_removal_operation z_patches;
+      change_summary z_patches;
       current_operation z_patches;
       W.scrollbox @@ current_hunks z_patches;
       Lwd.pure
--- a/lib/interactive_viewer.ml
+++ b/lib/interactive_viewer.ml
@@ -62,28 +62,33 @@ let navigate z_patches (dir : direction) : unit =
 
 let quit = Lwd.var false
 
-let calculate_add_remove lines =
-  List.fold_left
-    (fun (add, rem) line ->
-      match line with
-      | `Their _ -> (add + 1, rem)
-      | `Mine _ -> (add, rem + 1)
-      | _ -> (add, rem))
-    (0, 0) lines
-
-let modification_counter z_patches : ui Lwd.t =
+let calculate_add_remove_operation z_patches : ui Lwd.t =
   let$ z = Lwd.get z_patches in
   let p = Zipper.get_focus z in
-  let hunks = p.Patch.hunks in
   let additions, removals =
     List.fold_left
       (fun (add_acc, remove_acc) hunk ->
-        let addition, removal = calculate_add_remove hunk.Patch.lines in
+        let calculate_add_remove_hunk lines =
+          List.fold_left
+            (fun (addition, removal) line ->
+              match line with
+              | `Their _ -> (addition + 1, removal)
+              | `Mine _ -> (addition, removal + 1)
+              | `Common _ -> (addition, removal))
+            (0, 0) lines
+        in
+        let addition, removal = calculate_add_remove_hunk hunk.Patch.lines in
         (add_acc + addition, remove_acc + removal))
-      (0, 0) hunks
+      (0, 0) p.Patch.hunks
+  in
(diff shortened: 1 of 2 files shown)


lib/interactive_viewer.ml
```diff
@@ -52,20 +46,47 @@ let current_operation patches =
...
+  let$ current_patch = current_patch patches in
+  match current_patch with
+  | Some p ->
+      let hunks = p.Patch.hunks in
+      let additions, removals =
+        List.fold_left
+          (fun (add_acc, remove_acc) hunk ->
+            let addition, removal = calculate_add_remove hunk.Patch.lines in
+            (add_acc + addition, remove_acc + removal))
+          (0, 0) hunks
```
panglesd, line 68: Suggestion: move this and `calculate_add_remove` (renamed to `calculate_add_remove_hunk`) into a function named `calculate_add_remove_operation`. It makes it clearer what function does what!

lib/interactive_viewer.ml
```diff
@@ -52,20 +46,47 @@ let current_operation patches =
   | Some p -> ui_of_operation p.Patch.operation
   | None -> W.string "No operation"
 
+let calculate_add_remove lines =
+  List.fold_left
+    (fun (add, rem) line ->
+      match line with
+      | `Their _ -> (add + 1, rem)
+      | `Mine _ -> (add, rem + 1)
+      | _ -> (add, rem))
```
panglesd, line 55: It is good practice to try to have exhaustive match (without wildcards) as much as possible. This helps detect when there is a breaking change: for instance, the patch library add a 4th constructor. This 4th constructor would be included in the  `_` joker case, and the compiler would not help us detect that there is a change and that we need to handle it. With an "exhaustive" match, the addition of the 4th constructor would raise a "non-exhaustive match" warning and help us detect the change, so we can handle it!

panglesd (approved):
Nicely done!

I have left some comments in the code that require small changes, but it is very good in general.

panglesd:
Oh, and two more comments:

- It would be nice to handle the "plural": "1 addition", but "3 additions", and similarly "1 removal" but "5 removals"!
- You will need to rebase your PR on top of master since there is some unfortunate PR (due to ocamlformat...)

maha-sachin:
I have addressed all the above comments and had a little trouble with the latest code hence the delay. 
Do let me know if you have more comments. 

Also, the updated UI screenshot of the implementations is below 
Screenshot 2024-03-07 at 11 44 47 AM

lib/interactive_viewer.ml
```diff
@@ -60,9 +60,38 @@ let navigate z_patches (dir : direction) : unit =
   | Prev -> Lwd.set z_patches (Zipper.prev z)
   | Next -> Lwd.set z_patches (Zipper.next z)
 
-(* let pure_str s = Lwd.pure (W.string s) *)
 let quit = Lwd.var false
 
+let calculate_add_remove_operation z_patches : ui Lwd.t =
```
panglesd, line 65: Sorry if I wasn't clear! The organization in functions is not yet the one I had in mind.

We need to separate two distinct things:
- The computation of the two integers corresponding to the addition and removal
- and the creation of a ui element.

Ideally, the code for both should be in sufficiently distinct place, for instance two different function (or one function in another, but with a naming that suggests what it does.

But here, the `calculate_add_remove_operation` does not seem to return a UI element!

lib/interactive_viewer.ml
```diff
@@ -60,9 +60,38 @@ let navigate z_patches (dir : direction) : unit =
...
+  let$ z = Lwd.get z_patches in
+  let p = Zipper.get_focus z in
+  let additions, removals =
+    List.fold_left
+      (fun (add_acc, remove_acc) hunk ->
+        let calculate_add_remove_hunk lines =
```
panglesd, line 71: Including the function definition inside the `fold_left` is less readable and "redefines it" each time, in theory.

lib/interactive_viewer.ml
maha-sachin: @panglesd, Thank you for the clarification! I understand now. I'll work on separating the concerns into distinct functions:
1. One function will handle the computation of the integers corresponding to the addition and removal.
2. Another function will be responsible for creating the UI element based on those computed values.
I'll ensure that the code is organized in a way that reflects these separate responsibilities. Thank you for your patience!

maha-sachin:
Hi @panglesd, let me know what you think.

lib/interactive_viewer.ml
```diff
@@ -72,6 +109,7 @@ let view (patches : Patch.t list) =
   W.vbox
     [
       operation_info z_patches;
+      calculate_addition_removal_operation z_patches;
       current_operation z_patches;
```
panglesd, line 113: We have `operation_info`, `current_operation` and `current_hunks` here. The name `calculate_addition_removal_operation` is much longer and feels different: other names are descriptive of what they will display (they are ui components) while `calculate` seems to be an instruction. Try to make your name more similar to the one here! Ui component might need computation, but that's not how you describe them.

panglesd (commented):
Great, that's very close from being merged, but there are still some peculiar things!

- `count_lines_in_hunks` does not count the number of lines in a hunk. It take a _single_ line and an accumulator and adds the contribution of the line to the accumulator. So the name is very misleading!
- Similarly `calculate_addition_removal_in_patches` does not take multiple patches as input, but rather multiple `hunks`, so the name is also misleading.

Let me suggest some improvements!
- Define the functions that will never be useful outside of this context as local variables (for instance, the one you call `count_lines_in_hunk` should be local)
- Give the local functions simpler name. We already know the context since they are local, so they can be named, say, `add_hunk` (for instance) if they add the contribution of a hunk in the total number of additions and removal.
- Your code will look better with a `add_line` function.

maha-sachin:
@panglesd, Could you please take a look at the updates I made based on your suggestions? I've made several changes to improve the clarity and consistency of the function names. I would appreciate your feedback on the adjustments. Thank you

lib/interactive_viewer.ml
```diff
@@ -60,9 +60,40 @@ let navigate z_patches (dir : direction) : unit =
...
+    | `Common _ -> (addition, removal)
+  in
+  List.fold_left add_line (0, 0) lines
+
+let add_hunks hunks =
+  List.fold_left
+    (fun total_counts hunk ->
+      let additions, removals = add_lines hunk.Patch.lines in
+      (fun (add1, remove1) (add2, remove2) -> (add1 + add2, remove1 + remove2))
+        total_counts (additions, removals))
```
panglesd, line 79: You are creating an anonymous function that you are directly applying to some arguments.

This example is more obvious but very similar to yours:
```
(fun x y -> x + y) 4 10
```
is a convoluted version of
```
4 + 10
```

Could you change this similarly to avoid the unnecessary function?

panglesd (commented):
That's better!
There are still some things I'd like to improve. Functions that "add" something should take an accumulator (the number of additions/removal) and add the contribution of a line/hunk to this accumulator.
So, for instance, the function defined [line 66](https://github.com/panglesd/diffcessible/pull/30/files#diff-0b2d0a78372bb1d9e5c8972b915d555fee362e1cbd95e9a1d48ddf9a2812b5cbR66) is an "add" function and its name is right.
However, the functions defined line 65 and 74 are _not_ an "add" functions. They could be turned into "add" functions by giving an accumulator to add the contributions to. I believe you do want to add the accumulator to the function line 65, but maybe only rename the function line 65, eg to `additions_and_removal`!

maha-sachin:
> That's better! There are still some things I'd like to improve. Functions that "add" something should take an accumulator (the number of additions/removal) and add the contribution of a line/hunk to this accumulator. So, for instance, the function defined [line 66](https://github.com/panglesd/diffcessible/pull/30/files#diff-0b2d0a78372bb1d9e5c8972b915d555fee362e1cbd95e9a1d48ddf9a2812b5cbR66) is an "add" function and its name is right. However, the functions defined line 65 and 74 are _not_ an "add" functions. They could be turned into "add" functions by giving an accumulator to add the contributions to. I believe you do want to add the accumulator to the function line 65, but maybe only rename the function line 65, eg to `additions_and_removal`!

@panglesd, Would you prefer to rename add_hunk to something like accumulate_count or combine_hunks_counts to better reflect its purpose?

maha-sachin:
@Julow & @panglesd, Could you please look at the updates I made based on your suggestions? Thank you and I would appreciate your feedback on the adjustments.

lib/interactive_viewer.ml
```diff
@@ -60,9 +60,41 @@ let navigate z_patches (dir : direction) : unit =
...
+  let format_plural n singular plural =
+    if n = 1 then Printf.sprintf "%d %s" n singular
+    else Printf.sprintf "%d %s" n plural
+  in
+  let operation_count =
+    Printf.sprintf "%s, %s"
+      (format_plural total_additions "addition" "additions")
+      (format_plural total_removals "removal" "removals")
+  in
+  W.string ~attr:Notty.A.(fg lightcyan) @@ Printf.sprintf "%s\n" operation_count
```
Julow, line 96: This `\n` is used to render an empty line. Is it intended ?

If it is, I'd prefer it not to be implemented like this. Instead an empty line should be explicitly added in the `W.vbox` with `W.string "";`.

In any case, this should be simplified to:
```suggestion
  W.string ~attr:Notty.A.(fg lightcyan) operation_count
```

Julow (commented):
Looks very good !

I have a comment regarding the empty line. Otherwise, this is perfect and I would merge.

lib/interactive_viewer.ml
maha-sachin: @Julow, Thank you for your feedback! I'm glad to hear that you find the implementation satisfactory.
I'll address the comment regarding the empty line promptly. Once that's done, I'll let you know.

Julow: I agree the output looks better without the newline.

I do not see any new modification in the PR, did you push to the same branch ?

maha-sachin: @Julow wanted to check before making the change, the change should be in and ready to merge.

Julow (approved):
Perfect! Let's merge!

maha-sachin:
Thank you @Julow & @panglesd for merging the PR, @panglesd, for your detailed explanations and guidance throughout the process. Your insights were invaluable in understanding the concepts better. I appreciate your support and look forward to continuing to learn from @Julow and you.

PR #31, opened 2024-03-07
Clarify file type expected by Diffcessible

rmeis06:
Hello! I'm an outreachy applicant and was working on setting up diffcessible. In so doing I found the type of file needed by diffcessible wasn't clear in the README.md
I believe this PR makes the readme a little more clear, by giving an example of the kind of input file expected. And referencing that example.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,10 +21,15 @@ opam install . --deps-only --with-test
 
 ## Running the Project
 
-To run the project, you can use the following commands:
+This project runs using the output of a git diff command. Such as:
+```bash
+git diff > git.diff
+```
+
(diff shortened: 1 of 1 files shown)


panglesd (approved):
Thanks, very good idea! Maybe you could mention that there is already a file in the repo, `example.diff`, to test new features? See #28.

rmeis06:
> Thanks, very good idea! Maybe you could mention that there is already a file in the repo, `example.diff`, to test new features? See #28.

Sounds like a good idea! I've updated the PR to include it. Thanks!

PR #32, opened 2024-03-07
[yokurang]: Side-by-side diff view

yokurang:
Implemented something like this

image

We toggle using `t`. Any thoughts on the design?

@panglesd @Julow


panglesd:
That is a very good start!

Some early comments on the usability:
- It would be nice if the left and right areas were of always the same size: half of the available size
- The common lines should be aligned, or even better, not duplicated!
- We need to keep the `+` and `-` signs: the goal is to have an accessible interface, so we cannot rely on colors!

yokurang:
Hi,

Can I clarify on what you mean by "The common lines should be aligned, or even better, not duplicated!", as I thought a side by side view still show common lines from both files? Thanks

Julow:
Hi, sorry for the confusion. Yes the common lines should be duplicated at first. Other way of rendering this are for future issues.
However, the common lines should be aligned, for example this diff:

```diff
 FIRST
-a
+b
+b
 LAST
```

Should render to something like:

```
FIRST       FIRST
a           b
            b
LAST        LAST
```

In order to do this, it might be needed to make the `split_hunk` a bit smarter.

Perhaps by this approach: When you encounter a `` `Mine ``, collect all the `` `Mine `` that follows in a list, then the following `` `Their `` in an other list. If one list is smaller than the other, add empty lines to it.
If instead you encounter a `` `Their `` first, do the same but collect the `` `Their `` first then the `` `Mine ``.
The append these lists to `mine_acc` and `their_acc` as you are already doing.

You'd need to turn it into a recursive function instead of using a `fold_left`. You need to write a function that collects consecutive lines of a specified kind (`` `Mine `` or `` `Their ``). This function must return the remaining of the list to continue the iteration from the right place.

yokurang:
> Hello @yokurang !
> 
> Could you confirm that @Julow's explanation makes sense to you?
> 
> This PR is already pretty advanced!

Hi, yes, the explanation makes sense. Sorry for the delay, I have been quite busy with school work and some interviews. I'm finishing up soon, so I'll have time to finish this PR soon! Sorry for the delay.

yokurang:
Hi @Julow, I believe I modified `split_hunk` in the way you explained before. Could I ask you to verify is my solution is indeed correct?

Furthermore, @panglesd, I made the following progress:
1. I added + and - signs
2. I wanted to ask if you have any pointers for finding the total width of the viewbox so that I can calculate half the width and wrap the lines appropriately so the left and right areas are always of the same size. I just read up on `layout_width` from https://ocaml.org/p/nottui/0.3/doc/Nottui/Ui/index.html#val-layout_width. I haven't tried it yet, but wanted to ask if I am in the right direction. 
3. With regards to the point that common lines should be aligned, or even better, not duplicated, can I clarify if we are not duplicating common lines, does this mean that common lines should only be displayed on one side of the side-by-side diff view? On that note, I am also assuming that the following screenshot is wrong? 

image

Thank you

node_modules/.lsp/debug.log
Julow, line 1: This file shouldn't be committed!

lib/interactive_viewer.ml
```diff
@@ -167,12 +231,17 @@ let view (patches : Patch.t list) =
...
                   "Type 'h' to go to the help panel, 'q' to quit, 'n' to go to \
-                   the next operation, 'p' to go to the previous operation");
+                   the next operation, 'p' to go to the previous operation. Press 't' to toggle view mode.");
         ]
   in
   W.vbox [ ui ]
 
 let start patch = Ui_loop.run ~quit ~tick_period:0.2 (view patch)
+
+
```
Julow, line 247: You need to format with `dune build @fmt --auto-promote`.

lib/interactive_viewer.ml
```diff
@@ -104,6 +104,64 @@ let change_summary z_patches : ui Lwd.t =
...
+type view_mode = SideBySide | Normal
+
+let view_mode = Lwd.var SideBySide
+
+let toggle_view_mode () =
+  match Lwd.peek view_mode with
+  | Normal -> Lwd.set view_mode SideBySide
+  | SideBySide -> Lwd.set view_mode Normal
+
+let rec split_and_align_hunk hunks mine_acc their_acc =
```
Julow, line 119: There's a bug in the line numbers and in the aligning:
```
   1   (executable                             1   (executable
   2    (public_name diffcessible)             2    (public_name diffcessible)
   3    (name main)                            3    (name main)
   4 -  (libraries diffcessible cmdliner))     4   
   5                                           5 +  (libraries diffcessible cmdliner patch))
```

The added line is also line 4 !

lib/interactive_viewer.ml
```diff
@@ -104,6 +104,64 @@ let change_summary z_patches : ui Lwd.t =
...
+  match hunks with
+  | [] -> (List.rev mine_acc, List.rev their_acc)
+  | `Common _ as common :: t ->
+      split_and_align_hunk t (common :: mine_acc) (common :: their_acc)
+  | `Mine s :: t ->
+      split_and_align_hunk t (`Mine s :: mine_acc) (`Common "" :: their_acc)
+  | `Their s :: t ->
+      split_and_align_hunk t (`Common "" :: mine_acc) (`Their s :: their_acc)
+
+let lines_to_ui_with_numbers lines attr_line_number attr_change =
```
Julow, line 129: The line numbers complicate a bit the PR and are a great feature on their own. I'd like to make sure we have a solid implementation of the feature before we add line numbers.

What do you think of leaving line numbers out of this PR for now ? I've opened this issue: https://github.com/panglesd/diffcessible/issues/49

lib/interactive_viewer.ml
```diff
@@ -104,6 +104,64 @@ let change_summary z_patches : ui Lwd.t =
   in
   W.string ~attr:Notty.A.(fg lightcyan) operation_count
 
+
+(** Side by side diff view implementation **)
+
+type view_mode = SideBySide | Normal
+
+let view_mode = Lwd.var SideBySide
```
Julow, line 112: I think the default should be `Normal` for now.

Julow (commented):
This is awesome ! Happy to see such a high quality contribution.

> I wanted to ask if you have any pointers for finding the total width of the viewbox so that I can calculate half the width and wrap the lines appropriately so the left and right areas are always of the same size

[`Ui.resize`](https://ocaml.org/p/nottui/0.3/doc/Nottui/Ui/index.html#val-resize) should be used instead. `~w` and `~h` sets a fixed width and height, while `~sw` and `~sh` sets how the elements should share the remaining space, see [`layout_spec`](https://ocaml.org/p/nottui/0.3/doc/Nottui/Ui/index.html#type-layout_spec.sh).

> With regards to the point that common lines should be aligned, or even better, not duplicated, can I clarify if we are not duplicating common lines, does this mean that common lines should only be displayed on one side of the side-by-side diff view?

The common lines should be duplicated ! Sorry for the confusion.

node_modules/.lsp/debug.log
yokurang: Resolved

lib/interactive_viewer.ml
yokurang: Resolved.

lib/interactive_viewer.ml
yokurang: Ok. Resolved.

lib/interactive_viewer.ml
Julow: Still not formatted. If you use vscode, I suggest turning on formatting on save.

lib/interactive_viewer.ml
Julow: It seems that you didn't push your recent work.

yokurang: Hi @Julow, I haven't pushed it yet sorry 😅. I had to quickly work on my programming language design and implementation assignment before I got back to working on this PR. I can push my current progress tonight and hopefully finish by this Wednesday 👍

Julow: Ok, good luck with your assignment :)

lib/interactive_viewer.ml
```diff
@@ -111,6 +121,114 @@ let change_summary z_patches : ui Lwd.t =
...
+    (List.map
+       (fun line ->
+         let content, attr =
+           match line with
+           | `Common s -> (s, attr_line_number)
+           | `Mine s -> (s, attr_change)
+           | `Their s -> (s, attr_change)
+         in
+         let padded_content =
+           content ^ String.make (max_width - String.length content) ' '
```
Julow, line 193: What is the goal of this padding ?

It's currently misbehaving and adding a variable amount of unwanted empty lines between diff lines. Though, I wonder if there's a way not to use padding.

lib/interactive_viewer.ml
```diff
@@ -111,6 +121,114 @@ let change_summary z_patches : ui Lwd.t =
...
+  | (`Common _ as common) :: t ->
+      split_and_align_hunk t (common :: mine_acc) (common :: their_acc)
+  | `Mine s :: `Their s' :: t ->
+      split_and_align_hunk t (`Mine s :: mine_acc) (`Their s' :: their_acc)
+  | `Mine s :: t ->
+      split_and_align_hunk t (`Mine s :: mine_acc) (`Common "" :: their_acc)
+  | `Their s :: t ->
+      split_and_align_hunk t (`Common "" :: mine_acc) (`Their s :: their_acc)
+
+let wrap_text s =
```
Julow, line 147: Is wrapping needed to avoid having horizontal scrolling ? I think that horizontal scrolling is what we want in the end, so I'd suggest not to bother with line wrapping initially.

Currently, if the content is wider than the screen, it's truncated to the right. I think this is reasonable for the first version of the side-by-side view. Horizontal scrolling is something we'll want in the future, a lot of discussion happened in https://github.com/panglesd/diffcessible/issues/45

lib/interactive_viewer.ml
```diff
@@ -47,14 +47,24 @@ let ui_of_operation operation =
       Ui.hcat
         [ W.string "Modification of "; W.string ~attr:blue_bold_attr path ]
 
+let line_numbers_visible = Lwd.var false
```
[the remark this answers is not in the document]
yokurang: Removed attempts to toggle number line

lib/interactive_viewer.ml
```diff
@@ -215,3 +337,4 @@ let start_test patch events width height =
   Notty_unix.output_image init_image;
   Lwd.quick_release content_ui_root;
   print_newline ()
+
```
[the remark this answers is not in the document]
yokurang: I have `ocamlformat` applied to my `ocaml` file on saving. For context I am using NeoVim, and I have `null-ls` helping me manage auto-formatting on saving.

lib/interactive_viewer.ml
yokurang: I added padding initially as an attempt to make the spacing look better when text wrapping.. but since we want to forgo text wrapping at this moment I removed this (saved for later) as I don't think I'll be needing this without the text wrap function

lib/interactive_viewer.ml
yokurang: Got it, I'll forgo text wrapping for now. Thank you !

lib/interactive_viewer.ml
```diff
@@ -47,14 +47,7 @@ let ui_of_operation operation =
...
         [ W.string "Modification of "; W.string ~attr:blue_bold_attr path ]
 
-let ui_of_hunk hunk =
-  let line_to_string line =
-    match line with
-    | `Their text -> W.string ~attr:Notty.A.(fg red) text
-    | `Mine text -> W.string ~attr:Notty.A.(fg green) text
-    | `Common text -> W.string text
-  in
-  Ui.vcat @@ List.map line_to_string hunk.Patch.lines
```
Julow, line 57: This diff is an artifact of being based off https://github.com/panglesd/diffcessible/pull/57, which has been reverted since. Could you rebase or `merge main` ?

lib/interactive_viewer.ml
```diff
@@ -111,6 +104,65 @@ let change_summary z_patches : ui Lwd.t =
...
+  | Normal -> Lwd.set view_mode SideBySide
+  | SideBySide -> Lwd.set view_mode Normal
+
+let rec split_and_align_hunk hunks mine_acc their_acc =
+  match hunks with
+  | [] -> (List.rev mine_acc, List.rev their_acc)
+  | (`Common _ as common) :: t ->
+      split_and_align_hunk t (common :: mine_acc) (common :: their_acc)
+  | `Mine s :: `Their s' :: t ->
+      split_and_align_hunk t (`Mine s :: mine_acc) (`Their s' :: their_acc)
```
Julow, line 124: This pairs only one added line with one removed line. This might not be the output we expect, for example:
```
18 11     W.scrollbox
19 12       (W.vbox
20 13          [
21    -          pure_str "Hello world!";
22    -          string_of_counter counter_d;
   14 +          pure_str (string_of_operation patch.operation);
23 15            Lwd.pure
24 16            @@ Ui.keyboard_area
```

This diff renders like this in side-by-side:
```
  W.scrollbox                                                 W.scrollbox
    (W.vbox                                                     (W.vbox
       [                                                           [
         pure_str "Hello world!";
         string_of_counter counter_d;                                pure_str (string_of_operation patch.operation);
         Lwd.pure                                                    Lwd.pure
         @@ Ui.keyboard_area                                         @@ Ui.keyboard_area
```

It would be more intuitive if the whole block of removed line was paired with the added line.
Visually aligning the first removed line of the block with the first added line, leaving the empty line in the right side to be below the green line instead of above.

Julow (approved):
Looks good to me :)

I think the output could be further improved but if you prefer, that could be done in a later PR.

yokurang:
I think I'll leave that for a separate PR.. permission to merge sir? 🫡

Julow:
You can go ahead. I'd prefer a "Squash and merge" to make the history more regular.

Issue #18, opened 2024-02-28
Compute and display number of hunks in the status bar.


panglesd:
It would be nice to display the number of hunks for the current operation, in the top status bar.

For instance:

```
--- a/path/to/file
+++ b/path/to/file
```
could become:
```
--- a/path/to/file
+++ b/path/to/file
7 hunks
```

yokurang:
thanks @panglesd ! I think https://github.com/panglesd/diffcessible/issues/4 is a good go for me. Thank you for asking !

yokurang:
Hi @panglesd, I just finished a PR for https://github.com/panglesd/diffcessible/issues/4 and would like to get back to this one if possible the PR is ok. Is ok if I reserve this issue? On another note, this issue also seems similar to https://github.com/panglesd/diffcessible/issues/19.

Dezzy12:
Hello @Julow, Can I get this issue?

panglesd:
Hello @Dezzy12 !

In order to make sure not to block issues for too long, we require applicants to submit a screenshot of the help page of the tool before assigning them an issue, ensuring that they are able to build the project.

Once that is done, I'll be happy to assign you the issue!

iAlexeze:
Hello @panglesd 
I'd like to be assigned this issue.

Attached is my setup ready to take it on.

![Screenshot 2024-03-06 113909](https://github.com/panglesd/diffcessible/assets/126210538/0a2c60e0-2455-41d5-9480-6d512125f3b7)

iAlexeze:
Thank you.

panglesd:
@iAlexeze let you be assigned to this issue!

@Dezzy12 although you asked earlier to be assigned to this issue, we believe that now that you have completed a good first issue, you should pick a "medium issue". We will add more medium issues soon!

iAlexeze:
I have made a draft of implementation on this issue, and submitted a pull request.
Kindly review @panglesd.

PR #33, opened 2024-03-08
Display Number of Hunks in Interactive Viewer Status Bar

iAlexeze:
This PR introduces an enhancement to the interactive viewer's status bar, specifically to display the number of hunks for the current operation. This update aims to improve the user experience by providing a clearer overview of the changes being reviewed.

**Changes:**
- Modified the **_operation_ info_** function in **_interactive_viewer.ml_** to include the number of hunks in the status bar display.

**Purpose:**
The primary goal of this update is to make the interactive viewer more informative and user-friendly. 
By displaying the number of hunks, users can quickly understand the scope of changes for each operation, facilitating a more efficient review process.

@panglesd

### DIFF
--- a/lib/interactive_viewer.ml
+++ b/lib/interactive_viewer.ml
@@ -8,8 +8,8 @@ let operation_info z_patches : ui Lwd.t =
   let num_hunks = List.length p.Patch.hunks in
   let hunk_text =
     match num_hunks with
-    | n when n > 1 || n = 0 -> Printf.sprintf "%d hunks" n
-    | _ -> "1 hunk"
+    | 1 -> "1 hunk"
+    | _ -> Printf.sprintf "%d hunks" num_hunks
   in
   W.string
     ~attr:Notty.A.(fg lightcyan)
--- a/lib/interactive_viewer.ml
+++ b/lib/interactive_viewer.ml
@@ -6,12 +6,17 @@ let operation_info z_patches : ui Lwd.t =
   let$ z = Lwd.get z_patches in
   let p = Zipper.get_focus z in
   let num_hunks = List.length p.Patch.hunks in
+  let hunk_text = match num_hunks with
+    | 1 -> "1 hunk"
+    | n when n > 1 -> Printf.sprintf "%d hunks" n
+    | _ -> "0 hunks"
+in
   W.string
     ~attr:Notty.A.(fg lightcyan)
-    (Printf.sprintf "Operation %d of %d, Hunks: %d "
+    (Printf.sprintf "Operation %d of %d, %s "
        (Zipper.get_current_index z + 1)
        (Zipper.get_total_length z)
-       num_hunks)
+       hunk_text)
 
 let ui_of_operation operation =
   let green_bold_attr = Notty.A.(fg green ++ st bold) in
--- a/lib/interactive_viewer.ml
+++ b/lib/interactive_viewer.ml
@@ -4,11 +4,14 @@ open Lwd_infix
 
 let operation_info z_patches : ui Lwd.t =
   let$ z = Lwd.get z_patches in
+  let p = Zipper.get_focus z in
+  let num_hunks = List.length p.Patch.hunks in
   W.string
     ~attr:Notty.A.(fg lightcyan)
-    (Printf.sprintf "Operation %d of %d"
+    (Printf.sprintf "Operation %d of %d, Hunks: %d "
        (Zipper.get_current_index z + 1)
-       (Zipper.get_total_length z))
+       (Zipper.get_total_length z)
+       num_hunks)
 
 let ui_of_operation operation =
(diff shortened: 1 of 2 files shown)


panglesd (approved):
Very good!

The code has no issue, but I'd like to suggest a small change: I find the current display (e.g. `Hunks: 1` a bit misleading: It looks like I'm viewing the hunk number one (out of we don't know how many)).

Could you make it so that it displays `0 hunks`, `1 hunk`, `2 hunks`, ... ?

iAlexeze:
Yeah! I agree with you. I will do just that.
Thank you.

iAlexeze:
I have refined the hunk display logic for clarity.
Kindly review.
@panglesd

lib/interactive_viewer.ml
```diff
@@ -4,11 +4,19 @@ open Lwd_infix
 
 let operation_info z_patches : ui Lwd.t =
   let$ z = Lwd.get z_patches in
+  let p = Zipper.get_focus z in
+  let num_hunks = List.length p.Patch.hunks in
+  let hunk_text = match num_hunks with
+    | 1 -> "1 hunk"
+    | n when n > 1 -> Printf.sprintf "%d hunks" n
+    | _ -> "0 hunks"
```
panglesd, line 12: Those two cases could be simplified into one!

iAlexeze: Okay!
I have combined the cases to improve code readability and clarity.
@panglesd

iAlexeze: @panglesd and @Julow 
Could you kindly take a look at this implementation I made after review?

Thanks.

lib/interactive_viewer.ml
```diff
@@ -4,11 +4,18 @@ open Lwd_infix
...
+  let p = Zipper.get_focus z in
+  let num_hunks = List.length p.Patch.hunks in
+  let hunk_text = match num_hunks with
+    | n when n > 1 || n = 0 -> Printf.sprintf "%d hunks" n
+    | _ -> "1 hunk"
+in
   W.string
     ~attr:Notty.A.(fg lightcyan)
-    (Printf.sprintf "Operation %d of %d"
+    (Printf.sprintf "Operation %d of %d, %s "
```
Julow, line 15: The space at the end of the string seems unnecessary.

lib/interactive_viewer.ml
```diff
@@ -4,11 +4,18 @@ open Lwd_infix
 
 let operation_info z_patches : ui Lwd.t =
   let$ z = Lwd.get z_patches in
+  let p = Zipper.get_focus z in
+  let num_hunks = List.length p.Patch.hunks in
+  let hunk_text = match num_hunks with
```
Julow, line 9: You need to run autoformatting with `dune build @fmt --auto-promote` (or via the "format on save" feature of the ocaml vscode plugin). You need to install `ocamlformat.0.26.1` with Opam to do that.

Julow (commented):
The code looks great :) Just two comments and I'd merge.

Julow:
This fixes https://github.com/panglesd/diffcessible/issues/18

lib/interactive_viewer.ml
iAlexeze: Hi @Julow 
I have removed the unnecessary space after the string in the status bar and also formatted the code with the `dune build @fmt --auto-promote` command.

Feels good to learn this autoformatting.

Kindly review.

lib/interactive_viewer.ml
```diff
@@ -4,11 +4,19 @@ open Lwd_infix
 
 let operation_info z_patches : ui Lwd.t =
   let$ z = Lwd.get z_patches in
+  let p = Zipper.get_focus z in
+  let num_hunks = List.length p.Patch.hunks in
+  let hunk_text =
+    match num_hunks with
+    | n when n > 1 || n = 0 -> Printf.sprintf "%d hunks" n
+    | _ -> "1 hunk"
```
Julow, line 12: I just noticed that this match could be written this way:

```suggestion
    | 1 -> "1 hunk"
    | _ -> Printf.sprintf "%d hunks" n
```

Julow (approved):
Found an other possible improvement. The code looks very good!

lib/interactive_viewer.ml
iAlexeze: Oh wow!
This is great. Didn't know.
Thank you.

But I think `n` should be replace by `num_hunks` which is the variable that holds the hunks.
Check the attached image.
@Julow 

image

Julow: You are absolutely right!

iAlexeze: Okay then, I have implemented the change as recommended.
Kindly review.

Thank you
@Julow

Julow (approved):
Perfect, thanks!

Issue #15, opened 2024-02-28
Bind `j` and `k` to scrolling


panglesd:
Requires #14 to be done first.

Keyboard arrows are common to scroll, but so are `j` and `k` on TUIs. The goal of this task is to add the following bindings:

- `j` to scroll down,
- `k` to scroll up.

Keybindings can be either added in `interactive_viewer.ml` alongside the other key bindings, or in the keybindings defined in the `vscroll_area` function, in `nottui_widgets.ml`

maha-sachin:
Hi @panglesd , @Julow ,
I would like to work on this issue. Once #14 is done can you Please assign me this?

maha-sachin:
@panglesd , Thank you for the update! If any other tasks become available or are better suited for me, I'm flexible and open to being reassigned accordingly. Looking forward to contributing further to the project!

maha-sachin:
Hi @panglesd , @Julow ,
I would like to work on this issue. Once https://github.com/panglesd/diffcessible/issues/14 is done can you Please assign me this?

PR #36, opened 2024-03-10
keyboard arrow scrolling implementation (binds up and down key to scrolling)

### LINKED ISSUE #14: Use keyboard arrows for scrolling
Currently, it's only possible to scroll the visible part of the hunk via the mouse. This isn't a very accessible interface.

The goal of this task is to add the following key-bindings:
- `up`: Scrolls toward the top.
- `down`: Scrolls toward the bottom.

Scrolling is implemented by [Nottui_widgets](https://ocaml.org/p/nottui/latest/doc/Nottui_widgets/index.html). We are currently using `W.scrollbox` but we'd need to switch to `W.vscroll_area` to have the keybindings above.

`W.scrollbox` is a nottui widget that implements scrolling using the mouse. The "scrolling state" is hidden in the widget and we cannot add any shortcut to modify it. So, we cannot use this widget for our purpose.
`W.vscroll_area` is a widget that implements (vertical) scrolling using the keyboard.

So, solving this task requires to switch to `W.vscroll_area`. This function takes as input a `scroll_state` and a function to modify the scroll state, that need to be create. It will be handy to be able to modify the scroll state from the outside of the function to add new keybindings.

Siddhi-agg:
This PR would resolve #14 . @Julow Kindly review it.
I also wanted to ask: what should I implement in the change function of the vscroll_area.

### DIFF
--- a/lib/interactive_viewer.ml
+++ b/lib/interactive_viewer.ml
@@ -70,11 +70,8 @@ let view (patches : Patch.t list) =
     | None -> failwith "zipper_of_list: empty list"
   in
   let curr_scroll_state = Lwd.var W.default_scroll_state in
-  let change_scroll_state action state =
-    let off_screen =
-      (match action with `Action -> true | `Content -> true)
-      && state.W.position > state.W.bound
-    in
+  let change_scroll_state _action state =
+    let off_screen = state.W.position > state.W.bound in
     if off_screen then
       Lwd.set curr_scroll_state { state with position = state.W.bound }
     else Lwd.set curr_scroll_state state
--- a/lib/interactive_viewer.ml
+++ b/lib/interactive_viewer.ml
@@ -69,11 +69,29 @@ let view (patches : Patch.t list) =
     | Some z -> Lwd.var z
     | None -> failwith "zipper_of_list: empty list"
   in
+  let curr_scroll_state = Lwd.var W.default_scroll_state in
+  let change_scroll_state action state =
+    let off_screen = action = `Action && state.W.position > state.W.bound in
+    let scroll_on_output =
+      action = `Action
+      &&
+      let state' = Lwd.peek curr_scroll_state in
+      state'.W.position = state'.W.bound
+      && state.W.position = state'.W.position
+      && state.W.position < state.W.bound
+    in
+    if scroll_on_output || off_screen then
+      Lwd.set curr_scroll_state { state with position = state.W.bound }
+    else Lwd.set curr_scroll_state state
+  in
   W.vbox
     [
       operation_info z_patches;
       current_operation z_patches;
-      W.scrollbox @@ current_hunks z_patches;
+      W.vscroll_area
+        ~state:(Lwd.get curr_scroll_state)
+        ~change:change_scroll_state
+      @@ current_hunks z_patches;
       Lwd.pure
       @@ Ui.keyboard_area
            (function
--- a/lib/interactive_viewer.ml
+++ b/lib/interactive_viewer.ml
@@ -72,15 +72,8 @@ let view (patches : Patch.t list) =
   let curr_scroll_state = Lwd.var W.default_scroll_state in
   let change_scroll_state action state =
     let off_screen = action = `Action && state.W.position > state.W.bound in
-    let scroll_on_output =
-      action = `Action
-      &&
-      let state' = Lwd.peek curr_scroll_state in
-      state'.W.position = state'.W.bound
-      && state.W.position = state'.W.position
-      && state.W.position < state.W.bound
-    in
-    if scroll_on_output || off_screen then
+    let off_screen_content = action = `Content && state.W.position > state.W.bound in
+    if off_screen || off_screen_content then
       Lwd.set curr_scroll_state { state with position = state.W.bound }
     else Lwd.set curr_scroll_state state
   in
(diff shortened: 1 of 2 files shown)


lib/interactive_viewer.ml
```diff
@@ -69,11 +69,29 @@ let view (patches : Patch.t list) =
     | Some z -> Lwd.var z
     | None -> failwith "zipper_of_list: empty list"
   in
+  let curr_scroll_state = Lwd.var W.default_scroll_state in
+  let change_scroll_state action state =
+    let off_screen = action = `Action && state.W.position > state.W.bound in
+    let scroll_on_output =
```
Julow, line 75: What does `scroll_on_output` mean ? Why isn't `off_screen` enough to avoid scrolling too far ?

lib/interactive_viewer.ml
```diff
@@ -69,11 +69,29 @@ let view (patches : Patch.t list) =
     | Some z -> Lwd.var z
     | None -> failwith "zipper_of_list: empty list"
   in
+  let curr_scroll_state = Lwd.var W.default_scroll_state in
+  let change_scroll_state action state =
+    let off_screen = action = `Action && state.W.position > state.W.bound in
```
Julow, line 74: Your function do not handle the `` `Content `` action, which is called when the size of the content of the area changes. You need to handle that the `bound` can change in this case.

This creates a bug:
- Go to a long diff hunk (for example, the fourth file in `example.diff`)
- Scroll to the bottom
- Go to a shorter diff hunk (for example, the next one)
- The screen is empty because the scroll is now too far off.

lib/interactive_viewer.ml
```diff
@@ -69,11 +69,29 @@ let view (patches : Patch.t list) =
...
     [
       operation_info z_patches;
       current_operation z_patches;
-      W.scrollbox @@ current_hunks z_patches;
+      W.vscroll_area
+        ~state:(Lwd.get curr_scroll_state)
+        ~change:change_scroll_state
+      @@ current_hunks z_patches;
       Lwd.pure
       @@ Ui.keyboard_area
```
Julow, line 96: Now that you have `curr_scroll_state`, you can change it from a keybinding, which are defined in `Ui.keyboard_area`.

Julow (commented):
Congrat on going so far :) I've added some explanations in comments. Please answer if you have more questions.

lib/interactive_viewer.ml
Siddhi-agg: @Julow Thanks for explaining how the ```state``` and ```change``` function for ```vscroll_area``` work. That helped a lot!
I have added a commit based on what you said about handling ``` `Content``` in the ```change``` function. Please review.

> Now that you have `curr_scroll_state`, you can change it from a keybinding, which are defined in `Ui.keyboard_area`.

I noticed #15 is based on that, and since I am already working on the scrolling implementation and the issue is unassigned right now, I would like to work on it.

lib/interactive_viewer.ml
```diff
@@ -69,11 +69,24 @@ let view (patches : Patch.t list) =
     | Some z -> Lwd.var z
     | None -> failwith "zipper_of_list: empty list"
   in
+  let curr_scroll_state = Lwd.var W.default_scroll_state in
+  let change_scroll_state action state =
+    let off_screen = action = `Action && state.W.position > state.W.bound in
+    let off_screen_content =
+      action = `Content && state.W.position > state.W.bound
+    in
+    if off_screen || off_screen_content then
```
Julow, line 78: The exclusive `action = ...` checks (that cannot both be true at the same time) hints me at pattern matching:

```ocaml
  let off_screen =
    match action with
    | `Action -> ...
    | `Content -> ...
  in
```

I think that would make the code clearer, can you have a look ?

lib/interactive_viewer.ml
Julow: I'd prefer to assign #15 to someone that haven't worked with keybindings as you already shown that you can do that.

Julow (approved):
This works very nicely! I have just a suggestion to the code.

lib/interactive_viewer.ml
Siddhi-agg: Yeah! That would be a clearer implementation for these checks (because they cannot be both true simultaneously, as you said). I will get on making appropriate changes right away. Thanks for pointing it out!

Siddhi-agg:
@Julow I am not sure this is exactly how you wanted me to use pattern matching but this is one implementation of it. Please give a review.

lib/interactive_viewer.ml
Siddhi-agg: @Julow That makes sense! Then I am available for any other issues when they become available.

lib/interactive_viewer.ml
Julow: Sorry, I completely missed that the condition was the same in both cases !
The match always returns `true` in all branches so it can entirely be removed and replaced by `true`.
Then, `true && x` can be simplified to `x`.

Siddhi-agg: @Julow Yeah, I too did not notice this. Altho when I tried to do what you asked, I got the following error:
```
Error (warning 27 [unused-var-strict]): unused variable action.
```
I am stuck here, can you please help me?

Julow: It means that the variable `action` is not used within the function and that you should make that clear. We usually do that by adding a `_` at the beginning of unused variables:

```
let change_scroll_state _action state =
```

Julow (approved):
This is now perfect! Thanks!

Issue #9, opened 2024-02-22
Test for the interface


Julow:
A new entry point in the `Interactive_viewer` library should be added, which allows to:
- Run the UI in a dummy terminal not connected to a real terminal of a fixed size.
- Send a sequence of predefined events.
- Print the last frame.

This can then be wrapped in a binary and called from a cram test.

Shedrack-Sunday:
Hello @panglesd , I can't seem to find a first issue. Please can you assign this to me? I'll drop the screenshot of the helper page before the end of today. Thank you

panglesd:
Hello @Shedrack-Sunday !

We will be adding new first issues soon! Also, we won't assign an issue before the screenshot. Hopefully when you have the screenshot, there will be new issues available!

Sorry about that...

Dezzy12:
I think I might have an idea on what to do here, please can this issue be assigned to me?

panglesd:
Since you already have PRs merged, let this medium issue be assigned to you @Dezzy12 !

Julow:
> Run the UI in a dummy terminal not connected to a real terminal of a fixed size.

For clarity, by "dummy terminal" I'm not necessary talking about [creating a `Term.t`](https://ocaml.org/p/notty/latest/doc/Notty_unix/Term/index.html#val-create) of the library notty. This could be a good solution but I think getting the final image might be hard.

The `output` file descriptor can perhaps be opened from `/dev/null`. The `input` file descriptor requires more work as we need to write into it. See the [Unix](https://v2.ocaml.org/releases/5.1/api/Unix.html) module of the standard library.

An other solution is to write a different version of `Ui_loop.run`, defined in `vendor/lwd/lib/nottui/nottui.ml`, that do not use a `Term` but instead remember the last image using a `ref`. This means comming up with an implementation of `Term.pending` and `Term.event` that do not read a file descriptor but instead use a predefined list of events.

Dezzy12:
Hello @panglesd, @Julow 

I'm currently facing difficulties in proceeding with the assigned task of adding a new entry point to the Interactive_viewer library. To start, I created a `dummy_terminal.ml` file where I initialized the terminal. Additionally, I created an empty list to store the events, but I'm unsure which events need to be included. Also not sure if how to go about it

I would greatly appreciate your assistance in guiding me on how to proceed with this issue and how I can achieve an optimal solution.

Julow:
Hi @Dezzy12, sorry for the slow reply.

[...]

PR #42, opened 2024-03-12
Test for the interface

Dezzy12:
Hello,
I wanted to provide an update on my progress regarding the task mentioned in issue #9 . I have implemented a proposed solution for the issue and would appreciate a review.

@Julow @panglesd

### DIFF
--- a/lib/interactive_viewer.ml
+++ b/lib/interactive_viewer.ml
@@ -104,6 +104,15 @@ let change_summary z_patches : ui Lwd.t =
   in
   W.string ~attr:Notty.A.(fg lightcyan) operation_count
 
+let content_image = Lwd.var (Notty.I.vcat [ Notty.I.string Notty.A.empty "" ])
+
+let content_image_renderer (ui : ui) (height_request : int) =
+  let height = Ui.layout_height ui in
+  let width = Ui.layout_width ui in
+  let renderer = Renderer.make () in
+  Renderer.update renderer (width, height * height_request) ui;
+  renderer
+
 let view (patches : Patch.t list) =
   let help_panel =
     Ui.vcat
@@ -175,21 +184,23 @@ let view (patches : Patch.t list) =
   in
   W.vbox [ ui ]
 
-let input_events : Notty.Unescape.event list ref = ref []
-let output_image : Notty.image ref = ref Notty.I.empty
+  let content_length = List.length content in
+  let content_ui = W.vbox content in
+  let view_ui = Lwd.quick_sample (Lwd.observe content_ui) in
+  let image = Renderer.image (content_image_renderer view_ui content_length) in
+  Lwd.set content_image image;
+  content_ui
 
 let print_image () : unit =
-  let main_image = Notty.I.vcat [ Notty.I.string Notty.A.empty "" ] in
-  output_image := main_image;
-  Notty_unix.output_image ~fd:stdout !output_image
+  Notty_unix.output_image (Notty_unix.eol (Lwd.peek content_image))
 
 let start patch events =
   let rec loop () =
     if Lwd.peek quit then ()
     else (
-      input_events := events;
       print_image ();
       loop ())
   in
-  Ui_loop.run ~quit ~tick_period:0.2 (view patch);
-  loop ()
+  Ui_loop.run ~quit ~tick_period:0.2 (view patch events);
+  loop ();
+  print_image ()
--- a/lib/interactive_viewer.ml
+++ b/lib/interactive_viewer.ml
@@ -104,11 +104,6 @@ let change_summary z_patches : ui Lwd.t =
   in
   W.string ~attr:Notty.A.(fg lightcyan) operation_count
 
-let content_image_renderer (ui : ui) (width : int) (height : int) =
-  let renderer = Renderer.make () in
-  Renderer.update renderer (width, height) ui;
-  renderer
-
 let view (patches : Patch.t list) =
   let help_panel =
     Ui.vcat
@@ -181,29 +176,30 @@ let view (patches : Patch.t list) =
   W.vbox [ ui ]
 
 let start patch = Ui_loop.run ~quit ~tick_period:0.2 (view patch)
-let convert_char_to_key (c : char) : Ui.key = (`ASCII c, [])
 
 let start_test patch events width height =
+  let convert_char_to_key (c : char) : Ui.key = (`ASCII c, []) in
   let content_ui_root = Lwd.observe (view patch) in
   let content_ui = Lwd.quick_sample content_ui_root in
-  let ui_renderer = ref (content_image_renderer content_ui width height) in
+  let ui_renderer =
+    let renderer = Renderer.make () in
+    Renderer.update renderer (width, height) content_ui;
+    renderer
+  in
 
   let rec process_events (events : char list) =
     match events with
     | [] -> ()
     | event :: rest ->
         let ui_event = convert_char_to_key event in
-        ignore (Renderer.dispatch_key !ui_renderer ui_event);
-        Renderer.update !ui_renderer (width, height)
+        ignore (Renderer.dispatch_key ui_renderer ui_event);
+        Renderer.update ui_renderer (width, height)
           (Lwd.quick_sample content_ui_root);
         process_events rest
   in
 
   process_events events;
-  let init_image =
-    Renderer.image
-      (content_image_renderer (Lwd.quick_sample content_ui_root) width height)
-  in
+  let init_image = Renderer.image ui_renderer in
   Notty_unix.output_image init_image;
   Lwd.quick_release content_ui_root;
   print_newline ()
--- a/lib/interactive_viewer.ml
+++ b/lib/interactive_viewer.ml
@@ -187,23 +187,24 @@ let convert_char_to_key (c : char) : Ui.key = (`ASCII c, [])
 
 let start_test patch events =
   let content_ui_root = Lwd.observe (view patch) in
+  let content_ui = Lwd.quick_sample content_ui_root in
+  let ui_renderer = content_image_renderer content_ui in
 
-  let rec process_events (events : char list) (image : Notty.image) :
-      Notty.image =
+  let rec process_events (events : char list) =
     match events with
-    | [] -> image
+    | [] -> ()
     | event :: rest ->
         let ui_event = convert_char_to_key event in
 
-        let content_ui = Lwd.quick_sample content_ui_root in
-        let ui_renderer = content_image_renderer content_ui in
-
         ignore (Renderer.dispatch_key ui_renderer ui_event);
 
-        process_events rest (Renderer.image ui_renderer)
+        process_events rest
   in
 
-  let image = process_events events Notty.I.empty in
+  process_events events;
+  let init_image =
+    Renderer.image (content_image_renderer (Lwd.quick_sample content_ui_root))
+  in
+  Notty_unix.output_image init_image;
   Lwd.quick_release content_ui_root;
-  Notty_unix.output_image image;
   print_newline ()
--- a/lib/interactive_viewer.ml
+++ b/lib/interactive_viewer.ml
@@ -104,13 +104,11 @@ let change_summary z_patches : ui Lwd.t =
   in
   W.string ~attr:Notty.A.(fg lightcyan) operation_count
 
-let content_image = Lwd.var (Notty.I.vcat [ Notty.I.string Notty.A.empty "" ])
-
 let content_image_renderer (ui : ui) =
   let height = Ui.layout_height ui in
   let width = Ui.layout_width ui in
   let renderer = Renderer.make () in
-  Renderer.update renderer (width, height * 3) ui;
+  Renderer.update renderer (width, height * 4) ui;
   renderer
 
 let view (patches : Patch.t list) =
--- a/lib/interactive_viewer.ml
+++ b/lib/interactive_viewer.ml
@@ -106,11 +106,11 @@ let change_summary z_patches : ui Lwd.t =
 
 let content_image = Lwd.var (Notty.I.vcat [ Notty.I.string Notty.A.empty "" ])
 
-let content_image_renderer (ui : ui) (height_request : int) =
+let content_image_renderer (ui : ui) =
   let height = Ui.layout_height ui in
   let width = Ui.layout_width ui in
   let renderer = Renderer.make () in
-  Renderer.update renderer (width, height * height_request) ui;
+  Renderer.update renderer (width, height * 3) ui;
   renderer
 
 let view (patches : Patch.t list) =
@@ -184,23 +184,14 @@ let view (patches : Patch.t list) =
   in
   W.vbox [ ui ]
 
-  let content_length = List.length content in
-  let content_ui = W.vbox content in
-  let view_ui = Lwd.quick_sample (Lwd.observe content_ui) in
-  let image = Renderer.image (content_image_renderer view_ui content_length) in
+  let view_ui = Lwd.quick_sample (Lwd.observe (W.vbox content)) in
+  let image = Renderer.image (content_image_renderer view_ui) in
   Lwd.set content_image image;
-  content_ui
+  W.vbox content
 
 let print_image () : unit =
   Notty_unix.output_image (Notty_unix.eol (Lwd.peek content_image))
 
-let start patch events =
-  let rec loop () =
-    if Lwd.peek quit then ()
-    else (
-      print_image ();
-      loop ())
-  in
+let start patch events test =
   Ui_loop.run ~quit ~tick_period:0.2 (view patch events);
-  loop ();
-  print_image ()
+  if test then print_image ()
--- a/lib/interactive_viewer.ml
+++ b/lib/interactive_viewer.ml
@@ -104,11 +104,9 @@ let change_summary z_patches : ui Lwd.t =
   in
   W.string ~attr:Notty.A.(fg lightcyan) operation_count
 
-let content_image_renderer (ui : ui) =
-  let height = Ui.layout_height ui in
-  let width = Ui.layout_width ui in
+let content_image_renderer (ui : ui) (width : int) (height : int) =
   let renderer = Renderer.make () in
-  Renderer.update renderer (width, height * 4) ui;
+  Renderer.update renderer (width, height) ui;
   renderer
 
 let view (patches : Patch.t list) =
@@ -185,25 +183,26 @@ let view (patches : Patch.t list) =
 let start patch = Ui_loop.run ~quit ~tick_period:0.2 (view patch)
 let convert_char_to_key (c : char) : Ui.key = (`ASCII c, [])
 
-let start_test patch events =
+let start_test patch events width height =
   let content_ui_root = Lwd.observe (view patch) in
   let content_ui = Lwd.quick_sample content_ui_root in
-  let ui_renderer = content_image_renderer content_ui in
+  let ui_renderer = ref (content_image_renderer content_ui width height) in
 
   let rec process_events (events : char list) =
     match events with
     | [] -> ()
     | event :: rest ->
         let ui_event = convert_char_to_key event in
-
-        ignore (Renderer.dispatch_key ui_renderer ui_event);
-
+        ignore (Renderer.dispatch_key !ui_renderer ui_event);
+        let updated_content = Lwd.quick_sample content_ui_root in
+        ui_renderer := content_image_renderer updated_content width height;
         process_events rest
   in
 
   process_events events;
   let init_image =
-    Renderer.image (content_image_renderer (Lwd.quick_sample content_ui_root))
+    Renderer.image
+      (content_image_renderer (Lwd.quick_sample content_ui_root) width height)
   in
   Notty_unix.output_image init_image;
   Lwd.quick_release content_ui_root;
--- a/lib/interactive_viewer.ml
+++ b/lib/interactive_viewer.ml
@@ -182,14 +182,28 @@ let view (patches : Patch.t list) =
   in
   W.vbox [ ui ]
 
-  let view_ui = Lwd.quick_sample (Lwd.observe (W.vbox content)) in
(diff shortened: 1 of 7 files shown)


test/cram/interactive_viewer_dummy.ml
```diff
@@ -0,0 +1,56 @@
+open Cmdliner
+
+let input_events : Notty.Unescape.event list ref = ref []
+let output_image : Notty.image ref = ref Notty.I.empty
+
+let open_dummy_terminal () : unit =
+  let input_fd = Unix.openfile "/dev/null" Unix.[O_RDWR] 0o666 in
+  let output_fd = Unix.dup Unix.stdout in
```
Julow, line 8: I think this is inverted. We don't want to see stdout so it's the one that should be redirected to `/dev/null`. We do want to pass things to stdin, so it's the one that should be connected to a [pipe](https://v2.ocaml.org/releases/5.1/api/Unix.html#VALpipe).

test/cram/interactive_viewer_dummy.ml
```diff
@@ -0,0 +1,56 @@
...
+  Unix.dup2 input_fd Unix.stdin;
+  Unix.dup2 output_fd Unix.stdout;
+  Unix.close input_fd;
+  Unix.close output_fd
+
+
+let output_image : Notty.image = !output_image
```
Julow, line 15: This remember the very first image, which is empty. I think you mean to use `!output_image` in `print_image` instead of `output_image` and to remove this binding.

test/cram/interactive_viewer_dummy.ml
```diff
@@ -0,0 +1,56 @@
...
+
+let print_image () : unit =
+  Notty_unix.output_image output_image;
+  print_endline "Hello world";
```
Julow, line 19: Could you try putting this into the image with [`Notty.I.string`](https://ocaml.org/p/notty/latest/doc/Notty/I/index.html#val-string) ?

Julow (commented):
This is a very good start! Congrat on taking such a complex issue.

Here are the next steps:
- Write a [cram test](https://dune.readthedocs.io/en/stable/reference/cram.html) that runs the new executable.
- Print only the final image.
- Take inputs events from the command line.

Dezzy12:
I have implemented some of the requested changes.
@Julow

test/cram/test.t
```diff
@@ -0,0 +1,4 @@
+This is a cram test for the new executable.
+
+# Run the executable
+  $ ./dummy_terminal
```
Julow, line 4: The `./` should be removed.

test/cram/interactive_viewer_dummy.ml
```diff
@@ -4,18 +4,17 @@ let input_events : Notty.Unescape.event list ref = ref []
 let output_image : Notty.image ref = ref Notty.I.empty
 
 let open_dummy_terminal () : unit =
```
Julow, line 6: This doesn't really make sense at the moment. It doesn't have a function yet and prevents the executable from outputting anything. Let's comment it out for now: It might or might not be useful later.

Julow (commented):
Now the cram test runs with `dune runtest --auto-promote` and reflects the output of your executable.

However, it prints nothing.

Also, how to send input events to it ? I would like to write different invocations from the cram test that simulate different inputs.
The end goal is to have something like this:

```
Quit immediately, expect no output:
  $ dummy_terminal q

Navigate using 'n', expect to see the 4th diff:
  $ dummy_terminal n n n
```

The next step would be to accept inputs from the command line and to print the last image.
For now, just test sending `q`.

Dezzy12:
> However, it prints nothing.
Also, how to send input events to it ? I would like to write different invocations from the cram test that simulate different 

I noticed this and haven't found a way around it yet. Any help on how I can fix this?

Julow:
It prints nothing because `open_dummy_terminal` suppresses all output. I suggest removing `open_dummy_terminal` for now.

Dezzy12:
I have removed `open_dummy_terminal`. Somehow making it take inputs from the command line is still a challenge
@Julow

Julow:
The command line is defined using the [cmdliner](https://ocaml.org/p/cmdliner/latest/doc/index.html) library, [Arg.pos_all](https://ocaml.org/p/cmdliner/latest/doc/Cmdliner/Arg/index.html#val-pos_all) will take all the arguments into a list, use it instead of `const ()`.
Good luck ;)

Dezzy12:
I have made some changes such that it now accepts input from the command line

Dezzy12:
Fixed:
`(cram
 (deps %{bin:dummy_terminal}))`

Dezzy12:
Hello @Julow 
I face an issue while trying to run test with `opam exec -- dune runtest`

>  #Run the executable
   $ dummy_terminal q
  +Hello World
  +interactive_viewer_dummy: internal error, uncaught exception:
  +End_of_file          
  +[125]

Dezzy12:
I have done some changes. But the pull request check fails at the test level with the error:
`Error: Process completed with exit code 1.`

What next? @Julow

Dezzy12:
Yeah that's right. I have made come changes.

Julow:
Nice! It would be nice to send several inputs and see whether we still see only the last image.

test/cram/test.t
```diff
@@ -3,3 +3,9 @@ This is a cram test for the new executable.
 # Run the executable
   $ dummy_terminal q
   Hello World
+
+  $ dummy_terminal n n n
+  Hello World
+  Hello World
+  Hello World
+  Hello World
```
Julow, line 11: Ah, that's not what we want. We want the last image only!

Dezzy12:
I have fixed that.

Julow (commented):
Nice! We have the executable, the cram test and the dune setup.

Now we need a way to run the application into it. I mean run a modified version of `Interactive_viewer.start` that can accept custom inputs and return the last image in some way.
We currently call [`Ui_loop.run`](https://github.com/panglesd/diffcessible/blob/main/vendor/lwd/lib/nottui/nottui.ml#L852) to handle the loop. Perhaps we'd need a modified version of that too ? Or maybe switching to [`run_with_term`](https://github.com/panglesd/diffcessible/blob/main/vendor/lwd/lib/nottui/nottui.ml#L837) ?

Do you have an idea on how to go there ?

Dezzy12:
I am not sure of how to go about that. 
Some directions and guidance from you would be immensely helpful in getting me started.

Julow:
It's your job to come up with a plan. Start by reading and understanding the issue and the code I linked to.

Julow:
Congrat on the progress made so far!

Julow:
Can you explain what your last changes is doing ? Notably, how is it going to be used from the cram tests.

Dezzy12:
My last change includes the modification of the `Interactive_viewer.start` to make it accepts events(custom inputs), I then modified the interactive_viewer_dummy.ml to be able to run the application.
I was able to run test this using the command `./_build/default/test/cram/interactive_viewer_dummy.exe example.diff q` but I have not succeeded in running the cram test,  I encountered this error while running (`opam exec -- dune runtest`)

`#Run the executable
   $ dummy_terminal example.diff
+diffcessible: internal error, uncaught exception:
+Sys_error("example.diff: No such file or directory")	             
+[125]`

Julow:
It seems that the program is trying to open a file named `example.diff` but do not succeed. Try creating such a file in the cram test. [Doc about directory cram tests](https://dune.readthedocs.io/en/stable/reference/cram.html#file-and-directory-tests)

Dezzy12:
I created the file(example.diff) as part of the test, but while running the test it gets stuck at this level `Done: 99% (115/116, 1 left) (jobs: 1)` Does not go pass that.

Julow:
That might be because the program never terminate ? :thinking:

Dezzy12:
I thought so too, but I haven't figured it out yet. Perhaps I am missing something.

Also, I have a worry. The program normally does not terminate on its own until "q" is entered. This got me confused. Any tips on how to fix this?

lib/interactive_viewer.ml
```diff
@@ -77,7 +77,7 @@ let additions_and_removals lines =
     | `Mine _ -> (additions, removals + 1)
     | `Common _ -> (additions, removals)
   in
-  List.fold_left add_line (0, 0) lines
+  List.fold_left add_line (800, 600) lines
```
panglesd, line 80: Could you explain this change?

panglesd (commented):
You are getting closer to having a working test!

I think that you should now do a cleanup of the PR, before going further.
Try to make the changes minimal:
- Remove any unneeded addition from your PR.
- Remove the convoluted implementation (e.g., if you don't need a reference or Lwd variable, don't use one! If you don't need a loop, remove it!)

One general thing to keep in mind is that we don't want the addition of tests to modify the normal program. Currently, when hitting `q` in the normal program, some "image" is rendered on the terminal, which is something we want only for the test!

Once that is done, I'll give more precise instructions on how to go forward!

lib/interactive_viewer.ml
Dezzy12: This was a mistake

Dezzy12:
I have made some changes to the PR, please check @panglesd

lib/interactive_viewer.ml
```diff
@@ -103,34 +103,79 @@ let change_summary z_patches : ui Lwd.t =
...
+        Lwd.set quit true;
+        `Handled
+    | `ASCII 'n', [] ->
+        navigate z_patches Next;
+        `Handled
+    | `ASCII 'p', [] ->
+        navigate z_patches Prev;
+        `Handled
+    | _ -> `Unhandled
+  in
```
panglesd, line 146: Having a duplicated handling of events is not ideal! It would be better to have a list of `Ui.key` as the `events` argument.

panglesd (commented):
That's a little bit better!

Let me explain some concepts from Lwd.
- [`Notty.image`](https://ocaml.org/u/f3c72f25e5776ca3b4875324712542e3/notty/0.2.3/doc/Notty/index.html#type-image) are static rectangle of characters that makes "images"
- The [`Ui.t`](https://ocaml.org/p/nottui/latest/doc/Nottui/Ui/index.html#type-t) elements.They are a "superset" of images, as they can include more dynamic features, such as: handling focus, event, keys, ...
- [Renderers](https://ocaml.org/p/nottui/latest/doc/Nottui/Renderer/index.html). They are responsible for maintaining the state of the Ui, which mostly consists of dispatching events. From this state, you can get back the `Notty.image`.
- The [`Ui_loop`](https://ocaml.org/p/nottui/latest/doc/Nottui/Ui_loop/index.html) is responsible for the event and display loop (waiting for an event, dispatch it, get the new image, rebuild the image in the terminal, loop, ...)

The difference between the testing case and the normal case is that in the first one, we do not need the ui loop:
- In the normal case, we want to wait for user input, and we want to handle the printing of every new image in the terminal
- In the test case, we don't want to wait for user input (there is none, the events are pre-given through the command line) and we don't want the printing of every new image on the terminal: we only want the last one!

So I think you should now focus on making two clearly different paths for each case:
- In normal mode, the path is unchanged
- When testing, the Ui loop is not used, but a renderer is used to dispatch the events. When all events have been dispatched, the image is taken and printed.

Let me know if something is unclear!

Dezzy12:
Hello @panglesd Thank you for the info, I have done some changes please check

lib/interactive_viewer.ml
```diff
@@ -133,4 +142,37 @@ let view (patches : Patch.t list) =
...
 
-let start patch = Ui_loop.run ~quit ~tick_period:0.2 (view patch)
+let start patch events test =
+  if test then (
+    let z_patches : 'a Zipper.t Lwd.var =
+      match Zipper.zipper_of_list patch with
+      | Some z -> Lwd.var z
+      | None -> failwith "zipper_of_list: empty list"
+    in
+    let handle_event (event : Notty.Unescape.event) : Ui.may_handle =
```
Julow, line 152: It's counter-productive to repeat the key handler, instead we should use `(view patch)` without changing it.

Looking at Nottui's documentation, I found that `Renderer.dispatch_key renderer event` can send a key event to the UI. You already have a renderer created by `content_image_renderer` so this should be easy to use.

Julow (commented):
You've made awesome progress, congrats on figuring out this complex issue !

lib/interactive_viewer.ml
Dezzy12: @Julow , just adding this here for reference and clarity: 
```ocaml
let content_ui = Lwd.quick_sample (Lwd.observe (view patch)) in    
let ui_renderer = content_image_renderer content_ui in

let ui_events = (List.map (fun e -> convert_char_to_key e) events) in

List.iter ( function | event ->
begin
  let dispatch_res = Renderer.dispatch_key !ui_renderer event in
  
  match dispatch_res with
  | `Handled -> print_endline "Handled event";
  | `Unhandled -> print_endline "Unhandled event";

  ignore (dispatch_res)
end
```

I initially looked through the docs and how to use `Renderer.dispatch_key`. I must admit that it is a much more feasible and elegant :smile: option but while testing, I noticed that the events were not getting processed as expected. On printing out if the event was handled, I always get "Unhandled event" response. I have looked through the implementation of `Renderer.dispatch_key` and I am sure this is due to the ui not having a focus for the event - or the event listener created by `Ui.keyboard_area` not getting the event. I have tried several methods to set focus but I am not quite sure on how to go about it after several failed attempts. Please can I have some hint on this?

lib/interactive_viewer.ml
```diff
@@ -103,6 +103,15 @@ let change_summary z_patches : ui Lwd.t =
   in
   W.string ~attr:Notty.A.(fg lightcyan) operation_count
 
+let content_image = Lwd.var (Notty.I.vcat [ Notty.I.string Notty.A.empty "" ])
```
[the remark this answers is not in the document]
panglesd: Actually, a ref is not needed, a simple let binding should work!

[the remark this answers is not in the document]
Julow: Yes a ref is needed! We must record the current image at each loop step so we are left with the last image when the loop exits.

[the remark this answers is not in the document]
Julow: Sorry, @panglesd is right. We don't need a ref if we call `Renderer.image` only once, after the loop at exited. We also don't need this let binding.

Dezzy12:
Hi @Julow,

I've made the changes as requested earlier. However, I wasn't able to resolve the repeated key handler
 issue. It kept producing unexpected output. Perhaps you could provide some additional details on how I can approach this?

lib/interactive_viewer.ml
Julow: It's normal to have some events that are unhandled (the ui do not handle all the keys) but this code don't seem to contain an obvious bug. Could you push your latest change to the branch so I can have a look ?

test/cram/test.t
```diff
@@ -0,0 +1,18 @@
+This is a cram test for the new executable.
+
+# Run the executable
+  $ touch example.diff
+  $ echo "diff --git a/dir1/file.txt b/dir2/file.txt\nindex dfd0d1e..ce59064 100644\n--- a/dir1/file.txt\n+++ b/dir2/file.txt\n@@ -1 +1 @@\n-This is the original content.\n+This is the modified content." > example.diff
```
panglesd, line 5: This syntax adds `\n` inside the file, instead of new lines, in my setup.

Also, the diff on which this is tested is quite minimal. Maybe it would be better to start from the `example.diff` present at the root of the repo?
You would not need to copy it to the relevant directory, and add it to the dependencies of the `cram` test in the `dune` file.

lib/interactive_viewer.ml
Dezzy12: I have pushed my latest changes to this branch

lib/interactive_viewer.ml
```diff
@@ -134,3 +141,14 @@ let view (patches : Patch.t list) =
...
 
 let start patch = Ui_loop.run ~quit ~tick_period:0.2 (view patch)
+let convert_char_to_key (c : char) : Ui.key = (`Uchar (Uchar.of_char c), [])
+
+let start_test patch events =
+  let content_ui = Lwd.quick_sample (Lwd.observe (view patch)) in
+  let ui_renderer = content_image_renderer content_ui in
+  let ui_events = List.map (fun e -> convert_char_to_key e) events in
+  List.iter
+    (fun event -> ignore (Renderer.dispatch_key ui_renderer event))
```
panglesd, line 151: A problem here is that you dispatch the key, but do not update the content of the renderer. Since we are not using the UI loop, we need to handle the changes of the Lwd elements by hand.
So for each dispatched event, you need to update the content of the renderer.

lib/interactive_viewer.ml
```diff
@@ -134,3 +141,14 @@ let view (patches : Patch.t list) =
     ]
 
 let start patch = Ui_loop.run ~quit ~tick_period:0.2 (view patch)
+let convert_char_to_key (c : char) : Ui.key = (`Uchar (Uchar.of_char c), [])
+
+let start_test patch events =
+  let content_ui = Lwd.quick_sample (Lwd.observe (view patch)) in
```
panglesd, line 147: When you call `quick_sample`, you need to call `quick_release` when you want the variables to be updated again.

lib/interactive_viewer.ml
```diff
@@ -134,3 +141,14 @@ let view (patches : Patch.t list) =
     ]
 
 let start patch = Ui_loop.run ~quit ~tick_period:0.2 (view patch)
+let convert_char_to_key (c : char) : Ui.key = (`Uchar (Uchar.of_char c), [])
```
panglesd, line 144: Why `` `Uchar `` instead of `` Key `` ?

lib/interactive_viewer.ml
Julow: Additionally, you need to call `quick_sample` after every events to allow the changes to propagate in the UI.
See the function `run_with_term` in `vendor/lwd/lib/nottui/nottui.ml` line 837 that implements an event loop. The function `step` calls `Lwd.quick_sample` and `Renderer.update` and handles events.
You need to call `quick_release` only once, when you're done observing the root.

Dezzy12:
I have implement the requested changes.

test/cram/test.t
```diff
@@ -0,0 +1,20 @@
+This is a cram test for the new executable.
+
+# Run the executable
+  $ dummy_terminal example.diff n n
+  Operation 2 of 14, 1 hunk
```
panglesd, line 5: There seem to be a problem here: Since two `n` are given, we would expect the operation number to be `3` (zero `n` is "Operation 1", one `n` is "Operation 2", two `n` is "Operation 3").

lib/interactive_viewer.ml
```diff
@@ -134,3 +141,27 @@ let view (patches : Patch.t list) =
...
+
+  let rec process_events (events : char list) (image : Notty.image) :
+      Notty.image =
+    match events with
+    | [] -> image
+    | event :: rest ->
+        let ui_event = convert_char_to_key event in
+
+        let content_ui = Lwd.quick_sample content_ui_root in
+        let ui_renderer = content_image_renderer content_ui in
```
panglesd, line 157: You don't need to create a new renderer each new event!

panglesd (commented):
There are small bugs and the style can be improved, but we are really getting there! You have a first version of a test for the user interface, congrats!

The thing you should focus now is to fix the bug mentioned as a comment below.

test/cram/test.t
Dezzy12: This has been fixed.

lib/interactive_viewer.ml
Dezzy12: This too

panglesd:
Nice! Before going further in the review, it would be nice if you could rebase your PR on the newest changes on main!

Dezzy12:
Rebase completed

panglesd:
The rebase did not seem to work properly. The current PR now includes commit that aren't supposed to be included...

Dezzy12:
The rebase issue has been rectified.
@panglesd

test/cram/test.t
```diff
@@ -0,0 +1,20 @@
+This is a cram test for the new executable.
+
+# Run the executable
+  $ dummy_terminal example.diff n n
```
panglesd, line 4: You are not updating the renderer with the new content, so the `keyboard_area`'s are never updated. However, e.g. with the help panel, the set of keys change!

If you try `dummy_terminal example.diff h q` you will see the end result is different from if you do that in the real application.

lib/interactive_viewer.ml
```diff
@@ -104,6 +104,13 @@ let change_summary z_patches : ui Lwd.t =
   in
   W.string ~attr:Notty.A.(fg lightcyan) operation_count
 
+let content_image_renderer (ui : ui) =
+  let height = Ui.layout_height ui in
+  let width = Ui.layout_width ui in
+  let renderer = Renderer.make () in
+  Renderer.update renderer (width, height * 4) ui;
```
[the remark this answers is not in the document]
Dezzy12: Please can you provide more info on how I can go about this? I'm a little bit confuse?

Dezzy12:
I have done some changes regarding the issues mentioned above please review

panglesd:
Two comments!

- You should not create a new renderer each time you want to update the ui, just use the `Renderer.update` function on your renderer,
- You should not remove a test when you want to add one! While `h q` is worth testing, so is `n`, `n n`, `h`, ...

Dezzy12:
Using `Renderer.update` is the better option, I have made some changes

panglesd:
Great!

Now that the PR is finished in terms of functionality, could you do a "cleaning pass" on it? Since many things have been modified, some function might not be needed, some would be better local, ...
Try to make the PR the most straightforward: do not use a ref if not needed, avoid recreating a renderer if not needed. Try to modify everything you would ask to modify if you would review the PR!

Dezzy12:
I have done some clean up on the PR, Is this okay now?

Dezzy12:
Thank you very much @panglesd @Julow for the insights provided on this PR. I have learnt a lot

PR #47, opened 2024-03-20
help panel implementation

### LINKED ISSUE #11: Create a help panel
Add a panel that contains the list of keybindings with an explanation on what they do.

The panel would occupy on the whole screen, and its visibility would be toggled using a key (such as `h`).

Siddhi-agg:
Resolves #11. As advised by you, I have created a new PR because of the merge conflicts in the older one #35 .

### DIFF
--- a/lib/interactive_viewer.ml
+++ b/lib/interactive_viewer.ml
@@ -69,6 +69,7 @@ let navigate z_patches (dir : direction) : unit =
   | Next -> Lwd.set z_patches (Zipper.next z)
 
 let quit = Lwd.var false
+let help = Lwd.var false
 
 let additions_and_removals lines =
   let add_line (additions, removals) line =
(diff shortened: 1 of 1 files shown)


lib/interactive_viewer.ml
```diff
@@ -104,6 +105,16 @@ let change_summary z_patches : ui Lwd.t =
   W.string ~attr:Notty.A.(fg lightcyan) operation_count
 
 let view (patches : Patch.t list) =
+  let help_panel =
+    Ui.vcat
+      [
+        W.string "Help Panel:\n";
+        W.string "h:   Toggle the help panel";
```
panglesd, line 112: `h` does not actually _toggle_ the help panel. Instead, one has to use `q` to get out of the help panel.

I would prefer the "toggling" mechanism advertised than the "`q` to quit help panel" currently implemented!

panglesd (approved):
Other than the small comments below, this is very good! I'll merge as soon as you've addressed the comments.

lib/interactive_viewer.ml
Siddhi-agg: @panglesd In my last PR for the same issue, I was asked to set ```q``` for exiting the help panel by @Julow. So do you want me to set ```h``` for toggling the help panel and when the help panel is open, ```q``` to do nothing ?

panglesd: Oh, yes he did say it! Sorry for the disorder.
Let's keep what @Julow told you to do! However, let's have the description match. So `h` should be described as "opening" not "toggling" the help panel.

Siddhi-agg:
@panglesd I have implemented the changes you requested. Hope the PR is good now :)

panglesd (approved):
Thanks a lot for your contribution!

PR #52, opened 2024-03-28
Added horizontal scrolling functionality

maha-sachin:
Hi @Julow @panglesd , As per your suggestion, I started by implementing the horizontal scroll area functionality, which is working as expected. However, I'm now facing difficulties in combining both vertical and horizontal scroll areas within the same current hunk.

I've thoroughly tested each scroll area separately, and they work fine individually. However, when I attempt to integrate them into the same current hunk, I'm encountering issues.

I believe that understanding the correct approach to combine both vertical and horizontal scroll areas will greatly benefit the overall usability and functionality of the current hunk display.

Could you please provide some guidance or suggestions on how to effectively combine both scroll areas within the same current hunk? Any insights or examples you can provide would be immensely helpful in overcoming this hurdle.

Additionally, I've prepared a draft pull request with my current progress for your review. While it's not yet complete due to the challenge above, I'd appreciate any feedback or suggestions you may have on my hscroll_area implementation.I look forward to your guidance and feedback.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,4 +1,4 @@
-name: Builds, tests & co
+name: CI
 
 on:
   push:
@@ -15,18 +15,16 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        include:
-          - os: macos-latest
-            ocaml-compiler: "5.1"
-          - os: ubuntu-latest
-            ocaml-compiler: "5.1"
-          - os: windows-latest
-            ocaml-compiler: "4.14.0"
+        os: 
+          - macos-latest
+          - ubuntu-latest
+        ocaml-compiler: 
+          - "5.1"
 
     runs-on: ${{ matrix.os }}
 
     steps:
-      - name: Checkout tree
+      - name: Checkout code
         uses: actions/checkout@v4
 
       - name: Set-up OCaml ${{ matrix.ocaml-compiler }}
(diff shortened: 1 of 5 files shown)


panglesd:
Hello @maha-sachin !

That's a good start! Thanks for opening the draft PR!

I think you should not be modifying `vscroll_area`. It seems you still have some changes on this function, made to add horizontal scrolling, but we do want separate functions.

With regards to combining `vscroll_area` and `hscroll_area`, you'll simply define a `vscroll_area` which contains an `hscroll_area`, which contains the actual content (the current operation displayed in our case!)

My suggestion for you now would be to just clean up this PR, and rebase on top of the last changes on main!

maha-sachin:
> Hello @maha-sachin !
> 
> That's a good start! Thanks for opening the draft PR!
> 
> I think you should not be modifying `vscroll_area`. It seems you still have some changes on this function, made to add horizontal scrolling, but we do want separate functions.
> 
> With regards to combining `vscroll_area` and `hscroll_area`, you'll simply define a `vscroll_area` which contains an `hscroll_area`, which contains the actual content (the current operation displayed in our case!)
> 
> My suggestion for you now would be to just clean up this PR, and rebase on top of the last changes on main!

Thank you for your feedback and guidance on the draft PR. I apologize for the delayed response, as I have been feeling unwell recently. I will do it as soon as possible. Thank you for your understanding and support.

PR #53, opened 2024-03-29
Add key binding to scroll to the top

### LINKED ISSUE #46: Add keybinding to scroll to the top
#36 has added the ability to scroll.

It would be nice to have a keybinding to scroll back to the top of the displayed operation.

Feel free to ask for help or clarification when working on this issue!

Dezzy12:
fixes #46 
I have made some changes to solve issue #46, please review
@panglesd @Julow

### DIFF
--- a/lib/interactive_viewer.ml
+++ b/lib/interactive_viewer.ml
@@ -113,7 +113,7 @@ let view (patches : Patch.t list) =
         W.string "q:   Quit the diffcessible viewer";
         W.string "n:   Move to the next operation, if present";
         W.string "p:   Move to the previous operation, if present";
-        W.string "t:   Scroll back to the top of the displayed operation.";
+        W.string "g:   Scroll back to the top of the displayed operation.";
       ]
   in
   let z_patches : 'a Zipper.t Lwd.var =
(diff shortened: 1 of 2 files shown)


lib/interactive_viewer.ml
```diff
@@ -167,6 +168,10 @@ let view (patches : Patch.t list) =
                  | `ASCII 'h', [] ->
                      Lwd.set help true;
                      `Handled
+                 | `ASCII 't', [] ->
```
Julow, line 171: Why did you chose `t` ?
This is not consistent with keybindings that users might already know as in `less`, it's used to jump between tags and in [`tig`](https://github.com/jonas/tig/blob/master/doc/manual.adoc#default-keybindings), it's used to toggle the tree view.

Julow (commented):
This works perfectly ! I'm not a fan of the chosen key binding but it's otherwise very good.

lib/interactive_viewer.ml
Dezzy12: What about using `m` since we are returning to the top which in some sense is the main view. Or what do you suggest?

panglesd: I would suggest to use the same keybinding as `less`!

Dezzy12: so a `g` ?

Dezzy12:
I have changed the key binding used. please take a look!
@panglesd

Julow:
Merged, thanks!

Dezzy12:
Thank you

PR #54, opened 2024-04-03
Added line numbers to diffs

### LINKED ISSUE #50: Show line numbers in diffs
Showing line numbers in a diff file is tricky as each line can have a different number before and after the patch is applied.

See for example, this diff from the `example.diff` file:

```diff
diff --git a/bin/main.ml b/bin/main.ml
index e03c4a4..cb71f31 100644
--- a/bin/main.ml
+++ b/bin/main.ml
@@ -1,7 +1,9 @@
 open Diffcessible
 
 let main () =
-  Interactive_viewer.start ()
+  let s = In_channel.input_all In_channel.stdin in
+  let patch = Patch.to_diffs s in
+  Interactive_viewer.start patch
 
 open Cmdliner
 
````

The line numbers are start at `1` in both files, as indicated by the `@@` line, but then diverges due to there being more added lines than removed lines.
The line numbers are as follow:
```
1  1  open Diffcessible
2  2  
3  3  let main () =
4    -  Interactive_viewer.start ()
   4 +  let s = In_channel.input_all In_channel.stdin in
   5 +  let patch = Patch.to_diffs s in
   6 +  Interactive_viewer.start patch
5  7  
6  8  open Cmdliner
7  9  
```

### PR DESCRIPTION
**Pull request by Siddhi-agg**
This PR resolves #50.
I have made modifications in the Patch library to include line numbers when displaying hunks. I have also fixed the bug which used to show starting line of Mine hunk and Their hunk as 0 at the beginning of the hunk. Please review.

### DIFF
--- a/test/cram/test.t
+++ b/test/cram/test.t
@@ -5,12 +5,12 @@ This is a cram test for the new executable.
   Operation 1 of 14, 1 hunk
   1 addition, 1 removal
   Modification of bin/dune
-  @@ -0,4 +0,4 @@
-    (executable
-     (public_name diffcessible)
-     (name main)
-  -  (libraries diffcessible cmdliner))
-  +  (libraries diffcessible cmdliner patch))
+  @@ -1,4 +1,4 @@
+  1 1   (executable
+  2 2    (public_name diffcessible)
+  3 3    (name main)
+  4   -  (libraries diffcessible cmdliner))
+    4 +  (libraries diffcessible cmdliner patch))
   
   
   
@@ -26,17 +26,17 @@ This is a cram test for the new executable.
   Operation 2 of 14, 1 hunk
   3 additions, 1 removal
   Modification of bin/main.ml
-  @@ -0,7 +0,9 @@
-    open Diffcessible
-    
-    let main () =
-  -   Interactive_viewer.start ()
-  +   let s = In_channel.input_all In_channel.stdin in
-  +   let patch = Patch.to_diffs s in
-  +   Interactive_viewer.start patch
-    
-    open Cmdliner
-    
+  @@ -1,7 +1,9 @@
+  1 1   open Diffcessible
+  2 2   
+  3 3   let main () =
+  4   -   Interactive_viewer.start ()
+    4 +   let s = In_channel.input_all In_channel.stdin in
+    5 +   let patch = Patch.to_diffs s in
+    6 +   Interactive_viewer.start patch
+  5 7   
+  6 8   open Cmdliner
+  7 9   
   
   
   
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: vendor/patch/src/patch.ml [hunk 1]
```diff
@@ -52,18 +52,31 @@ type hunk = {
 }
 
 let unified_diff hunk =
+  let left_line_num = ref hunk.their_start in 
+  let right_line_num = ref hunk.mine_start in
+  let incr ref =
```
Julow, line 57: This function already exists in the standard library and is called exactly like that.

vendor/patch/src/patch.ml
```diff
@@ -52,18 +52,31 @@ type hunk = {
...
+         | `Mine line ->
+            incr left_line_num;
+            (Printf.sprintf "%2d    - %s" !left_line_num line))
        hunk.lines)
 
 let pp_hunk ppf hunk =
   Format.fprintf ppf "@@@@ -%d,%d +%d,%d @@@@\n%s"
-    hunk.mine_start hunk.mine_len hunk.their_start hunk.their_len
+    (if hunk.mine_len = 0 then 0 else hunk.mine_start+1) hunk.mine_len 
+    (if hunk.their_len = 0 then 0 else hunk.their_start+1) hunk.their_len
```
Julow, line 79: Nice catch!

vendor/patch/src/patch.ml
```diff
@@ -52,18 +52,31 @@ type hunk = {
 }
 
 let unified_diff hunk =
+  let left_line_num = ref hunk.their_start in 
+  let right_line_num = ref hunk.mine_start in
```
Julow, line 56: I think these two are swapped. `right_line_num` is initialized with `hunk.mine_start` but incremented on each `` `Their ``.

I suggest using "mine" and "their" in the variable name to avoid confusion.

test/cram/test.t
```diff
@@ -47,11 +47,11 @@ This is a cram test for the new executable.
   Operation 3 of 14, 1 hunk
   1 addition, 1 removal
   Modification of lib/dune
-  @@ -0,3 +0,3 @@
-    (library
-     (name diffcessible)
-  -  (libraries notty nottui lwd))
-  +  (libraries notty nottui lwd patch))
+  @@ -1,3 +1,3 @@
```
Julow, line 50: It would be nice to have a test case showing a hunk that do not start at line 1 and where "mine" and "their" counts do not start at the same number.
This could be added to as a new .diff file containing just that and shown in this cram test.

Julow (commented):
Very nice!

test/cram/test.t
Siddhi-agg: @Julow Sure! I would read up on .diff files and start working on it. Do you want just one testcase for that in the new .diff file?

Julow: One testcase would be enough but if you find several testcases that are interesting for different reasons, add them too :)

Siddhi-agg: @Julow I have added two test cases: one where the hunk doesn't start at line 1, and one where the hunk doesn't start at line 1 and the "mine" and "their" counts don't start at the same line. Please review them.

Julow (approved):
The new testcases look very nice!
Let's merge, thanks!

PR #56, opened 2024-04-04
Updated the implementation for hscroll area

maha-sachin:
@panglesd @Julow , I've encountered a bug in the scroll area implementation and would appreciate your guidance in resolving it. 

When using a vscroll area inside hscroll area, I've noticed that the horizontal scrolling behaves correctly with the left and right keys. However, there's an issue with vertical scrolling – when using the up and down keys, the behavior is unexpected, and vice versa.
Thank you very much for your assistance and support in tackling this issue. Your guidance is invaluable in helping me improve my skills and contribute effectively to the project.


lib/interactive_viewer.ml
```diff
@@ -149,9 +156,19 @@ let view (patches : Patch.t list) =
...
+             @@ current_hunks z_patches; *)
+          W.hscroll_area
+            ~state:(Lwd.get curr_scroll_state_h)
+            ~change:change_scroll_state_h
+          @@ W.vscroll_area
+               ~state:(Lwd.get curr_scroll_state)
+               ~change:change_scroll_state
           @@ current_hunks z_patches;
           Lwd.pure
           @@ Ui.keyboard_area
```
Julow, line 174: > What are your thoughts on incorporating horizontal scroll keys v and b?

Key bindings would be a great addition ! In what other tools are `v` and `b` used for scrolling ? As we plan to have `j` and `k` for vertical scrolling, I'd suggest `h` and `l` to be consistent in the vim-like keybindings.

Julow, line 174: > When using a vscroll area inside hscroll area, I've noticed that the horizontal scrolling behaves correctly with the left and right keys. However, there's an issue with vertical scrolling – when using the up and down keys, the behavior is unexpected, and vice versa.

Can you describe the unexpected behavior and how to get to it ?

I notice one unexpected behavior:
On a hunk that is big enough to scroll both vertically and horizontally, the right key has no effect when pressed but when pressing the down key, update both the vertical and horizontal scroll to the expected position.

I'm not sure why this happens by reading the code. Could you try printing the values of the states. Here's a setup that can work with two terminals: In one terminal run the app with `dune exec -- diffcessible example.diff 2> stderr`, in the second terminal read this file with `tail -f stderr`. You can then interact with the app and see debug printing at the same time. Make sure to print to stderr, for example with `Printf.eprintf`.
Good luck!

Julow (commented):
Nice work! I'm also not sure why the horizontal scroll misbehave. Could you attempt at debugging it ?
I've opened several threads for separate discussion.

lib/interactive_viewer.ml
maha-sachin: Hi @Julow, I understand that we currently use h for opening the help panel. we may need to explore alternative options for incorporating horizontal scrolling keys while ensuring consistency and usability in our application. What are your thoughts on this?

Julow: Good catch! Then `v` and `b` seem like good options.

lib/interactive_viewer.ml
```diff
@@ -122,12 +122,19 @@ let view (patches : Patch.t list) =
...
   in
   let curr_scroll_state = Lwd.var W.default_scroll_state in
+  let curr_scroll_state_h = Lwd.var W.default_scroll_state in
   let change_scroll_state _action state =
     let off_screen = state.W.position > state.W.bound in
     if off_screen then
       Lwd.set curr_scroll_state { state with position = state.W.bound }
     else Lwd.set curr_scroll_state state
   in
+  let change_scroll_state_h _action state =
```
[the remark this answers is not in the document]
maha-sachin: > The existing scroll related function and variable could be renamed to have `_v` to make things clear.

I agree that clear and descriptive names are essential for maintaining readability.

lib/interactive_viewer.ml
maha-sachin: > Can you describe the unexpected behavior and how to get to it ?
> 
> I notice one unexpected behavior: On a hunk that is big enough to scroll both vertically and horizontally, the right key has no effect when pressed but when pressing the down key, update both the vertical and horizontal scroll to the expected position.
> 
> I'm not sure why this happens by reading the code. Could you try printing the values of the states. Here's a setup that can work with two terminals: In one terminal run the app with `dune exec -- diffcessible example.diff 2> stderr`, in the second terminal read this file with `tail -f stderr`. You can then interact with the app and see debug printing at the same time. Make sure to print to stderr, for example with `Printf.eprintf`. Good luck!

I'll investigate further and try to understand why the right key is not having any effect, I'll set up the suggested environment with two terminals as you described.
Once I have more information from the debug prints, I'll analyze the behavior and work on resolving the issue accordingly.

Thanks for your guidance, and I'll keep you updated on my progress!

Issue #5, opened 2024-02-22
Add colors to output


panglesd:
Even if they are not visible by everyone, colors help to read a diff.

It would be nice to add green and red background colors to the added and removed lines, in the output.

ademolaomosanya:
hello i'm an outreachy applicant and i would love to work on this task i already have the repo running on my local machine

yokurang:
Hi, I have a fork of the machine on locally and would love to implement this feature. Here is the screenshot as requested: 
![image](https://github.com/panglesd/diffcessible/assets/114996150/5e31c2ee-798c-421b-b25c-b4a868990d7d)

Julow:
Hi @ademolaomosanya,

Thanks for your enthusiasm and congrat for having everything setup :)

The issue is your's. Hunks are currently printed using the `Patch` library's printing function, called [from here](https://github.com/panglesd/diffcessible/blob/3d2cc0eb3373fc06b2651fc4d25c1822ba2877de/lib/interactive_viewer.ml#L13).

To be able to add colors, you might need to rewrite this function, in interactive_viewer.ml, and then add the colors.
Modifying the `Patch` library for this is not the best solution as the library is intended to be used for other purpose, while colors are only useful in this project.

Julow:
Hi @yokurang,

This issue have been assigned just before you requested it. Is here an other issue that would interest you ?
I can suggest this one, for improving the status bar: https://github.com/panglesd/diffcessible/issues/18

JeanFelixHAKIZIMANA:
can you suggest me another issue other than this one please?

Siddhi-agg:
Hi @ademolaomosanya! I noticed you have been inactive for a while, so are you still working on this issue?

maha-sachin:
@Julow, I've completed the implementation for issue #17 "Compute and display number of added and removed lines in status bar" , if @ademolaomosanya is unavailable, could you assign this issue to me for further enhancements?

Julow:
@Siddhi-agg @JeanFelixHAKIZIMANA Sorry, the project was more popular than we expected and are running out of issues. Unfortunately, we are not assigning new contributors :/

@maha-sachin I do not wish to unassign an issue so fast. Let's review #17 and see what's the next best step.

maha-sachin:
> @Siddhi-agg @JeanFelixHAKIZIMANA Sorry, the project was more popular than we expected and are running out of issues. Unfortunately, we are not assigning new contributors :/
> 
[...]

PR #57, opened 2024-05-19
added colours to output:

yokurang:


yokurang:
PR to address [issue #5](https://github.com/panglesd/diffcessible/issues/5)

lib/interactive_viewer.ml
```diff
@@ -47,7 +47,14 @@ let ui_of_operation operation =
       Ui.hcat
         [ W.string "Modification of "; W.string ~attr:blue_bold_attr path ]
 
-let string_of_hunk = Format.asprintf "%a" Patch.pp_hunk
+let string_of_hunk hunk =
```
panglesd, line 50: Small nitpick: this function is not anymore converting to a string but to a UI element, so the `ui_of_hunk` name would make more sense!

panglesd (approved):

panglesd:
The tests in CI should be fixed on another PR, and then this one rebased!

PR #60, opened 2024-06-09
added line numbers

yokurang:


lib/interactive_viewer.ml
```diff
@@ -127,32 +127,88 @@ let rec split_and_align_hunk hunks mine_acc their_acc =
...
-      in
-      W.string ~attr content)
+let lines_with_numbers lines attr_change line_num prefix =
+  List.mapi
+    (fun index line ->
+      let is_last = index + 1 = List.length lines in
+      match line with
+      | `Common s when s = "" && is_last -> ("", Notty.A.empty)
+      | `Mine s when s = "" && is_last -> ("", Notty.A.empty)
+      | `Their s when s = "" && is_last -> ("", Notty.A.empty)
```
panglesd, line 137: Why all those special casing here? When I read a diff, I'd like to know when there is an empty last line!

lib/interactive_viewer.ml
```diff
@@ -127,32 +127,88 @@ let rec split_and_align_hunk hunks mine_acc their_acc =
...
+      | `Common s ->
+          incr line_num;
+          (Printf.sprintf "%3d   %s" !line_num s, Notty.A.empty)
```
panglesd, line 140: So, if there are more than `999` lines, things start to be unaligned? That would be unfortunate!

lib/interactive_viewer.ml
```diff
@@ -127,32 +127,88 @@ let rec split_and_align_hunk hunks mine_acc their_acc =
...
+let lines_with_numbers lines attr_change line_num prefix =
+  List.mapi
+    (fun index line ->
+      let is_last = index + 1 = List.length lines in
+      match line with
+      | `Common s when s = "" && is_last -> ("", Notty.A.empty)
+      | `Mine s when s = "" && is_last -> ("", Notty.A.empty)
+      | `Their s when s = "" && is_last -> ("", Notty.A.empty)
+      | `Common s ->
+          incr line_num;
```
panglesd, line 139: Let's avoid references!

In any case, this `incr line_num` on all `` `Common | `Mine | `Their `` cases is dubious.

Why not just reuse the values of `mine_length` and `their_length` available in the patch record?

panglesd (commented):
A very good start!

Below a first round of review.

lib/interactive_viewer.ml
Julow: I think there's a bug with this line num computation, here's what I get:
```
@@ -1,7 @@                                    @@ +1,9 @@
  1   open Diffcessible                         1   open Diffcessible
  2                                             2   
  3   let main () =                             3   let main () =
  4 -   Interactive_viewer.start ()             4 +   let s = In_channel.input_all In_chan
  5                                             5 +   let patch = Patch.to_diffs s in
  6                                             6 +   Interactive_viewer.start patch
  7                                             7   
  8   open Cmdliner                             8   open Cmdliner
```
In the left part, lines 5 and 6 are wrong. Here are the correct line numbers:
```
@@ -1,7 +1,9 @@
 1  1   open Diffcessible
 2  2   
 3  3   let main () =
 4    -   Interactive_viewer.start ()
    4 +   let s = In_channel.input_all In_channel.stdin in
    5 +   let patch = Patch.to_diffs s in
    6 +   Interactive_viewer.start patch
 5  7   
 6  8   open Cmdliner
 7  9   
```

PR #61, opened 2024-06-09
pager shell

yokurang:


Julow (commented):
What's the usecase of this script ? Is it for testing ?

In any case, it's not a pager because I can't do `git diff | diffcessible` or `cat example.diff | diffcessible`. For that, the main function must be changed to listen on stdin if no filename are given as filename.

yokurang:
> What's the usecase of this script ? Is it for testing ?
> 
> 
> 
> In any case, it's not a pager because I can't do `git diff | diffcessible` or `cat example.diff | diffcessible`. For that, the main function must be changed to listen on stdin if no filename are given as filename.

Just for testing !

difftool-dune.sh
```diff
@@ -0,0 +1,23 @@
...
+echo "Arguments received: $@"
+file="$2"
+if [ -z "$file" ]; then
+    file="$1"
+fi
+echo "File to diff: $file"
+
+# Check if file exists in the specified location or in the current directory
+if [ ! -f "$file" ]; then
+    if [ -f "$(basename "$file")" ]; then
```
Julow, line 13: Can you give an example that require this ?
Also, diffcessible can do its own error checks, no need to add more.

difftool-dune.sh
```diff
@@ -0,0 +1,23 @@
...
+        file="$(basename "$file")"
+    else
+        echo "File does not exist: $file"
+        exit 1
+    fi
+fi
+
+echo "Running: dune exec -- diffcessible $file"
+dune exec -- diffcessible "$file"
```
Julow, line 22: It's not possible to pass options to the tool that way. What do you think of doing `dune exec -- diffcessible "$@"` ? That would pass every arguments without looking at them, making sure we get output from the tool.

PR #65, opened 2024-06-30
Add colours to output normal and refactored code

yokurang:


yokurang:
Here I wanted to add refactoring in this PR because I also wanted to include the new functions for normal mode when refactoring

yokurang:
I am also running into this particular error when trying to generate the `mli` files:

```
❯ 
❯ ocamlc -c -i lib/InteractiveViewer.ml
File "lib/InteractiveViewer.ml", line 1, characters 5-11:
1 | open Nottui
         ^^^^^^
Error: Unbound module Nottui
```

Was wondering if anyone might know how to resolve this? I made sure the `dune` file has `nottui` and that I have the package installed using `opam`.

panglesd (commented):
PascalCasing is not very idiomatic in OCaml... But refactoring in multiple module is a very good idea!

A few comments below.

lib/interactive_viewer.ml
```diff
@@ -104,6 +98,68 @@ let change_summary z_patches : ui Lwd.t =
...
+  let their_line_num = ref hunk.Patch.their_start in
+  let mine_line_num = ref hunk.Patch.mine_start in
+  let lines_ui =
+    List.map
+      (function
+        | `Common line ->
+            incr their_line_num;
+            incr mine_line_num;
+            W.string ~attr:Notty.A.empty
+              (Printf.sprintf "%2d %2d   %s" !mine_line_num !their_line_num line)
```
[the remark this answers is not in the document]
yokurang: Created new issue: https://github.com/panglesd/diffcessible/issues/66

dune-project
```diff
@@ -7,9 +7,9 @@
 (source
  (github panglesd/diffcessible))
 
-(authors "Paul-Elliot <peada@free.fr>" "Jules Aguillon <jules@j3s.fr>")
+(authors "Paul-Elliot <peada@free.fr>" "Jules Aguillon <jules@j3s.fr>" "Alan Matthew <alan.matthew.yk@gmail.com>")
```
Julow, line 10: Thanks for this!

Julow (commented):
- I agree with the `.mli`. These modules have a complex implementation, which should be hidden behind a documented and abstracted interface. 
- The casing feels non-idiomatic for OCaml but why not. I won't enforce my favorite casing.

lib/HelpView.mli
```diff
@@ -0,0 +1,13 @@
+(** This module provides UI components and attributes for a help panel in a Nottui-based application. *)
+
+val help_visible : bool Lwd.var
+(** [help_visible] is a variable indicating whether the help panel is currently visible. *)
+
+val help_panel : Nottui.ui
+(** [help_panel] constructs the user interface for the help panel. This includes various help-related commands and their corresponding visual attributes, making it easy for users to understand available interactions. *)
+
+val help_keyboard_area : Nottui.ui
+(** [help_keyboard_area] provides a keyboard interaction area where pressing 'q' will close the help panel. *)
```
panglesd, line 10: This value is not used.

The problem would not be solved by simply removing the value. In my opinion, the problem is more that the logic for opening/closing the help panel is spread between two very different location: `HelpView.ml` and `InteractiveViewer.ml`. This logic should be in one place!

lib/InteractiveViewer.mli
```diff
@@ -0,0 +1,21 @@
...
+    @return Executes indefinitely, handling UI events until termination conditions are met (e.g., user quits). *)
+
+val start_test : Patch.t list -> char list -> int -> int -> unit
+(** [start_test patch events width height] facilitates automated UI testing by simulating keypress events and rendering the UI to the terminal.
+    This function is useful for automated or manual testing of UI behaviors in response to specific inputs.
+    @param patch A list of patches to be displayed in the test.
+    @param events A list of characters representing simulated keypresses to interact with the UI.
+    @param width The width of the rendering area in characters.
+    @param height The height of the rendering area in characters.
+    @return Outputs the resulting UI state to the terminal, providing visual feedback on the UI's response to the simulated events. *)
```
panglesd, line 21: The return ocamldoc tag (`@return`) is usually used to explain the returned value of a function. In this case, the return value is unit.
The explanation for the side effect (outputing the UI state to the terminal) could be done outside of the tag. But I'm still fine with it like that!

lib/InteractiveViewer.mli
```diff
@@ -0,0 +1,21 @@
+(** This module defines the main user interface and interaction logic for a diff viewer application using the Nottui library. It provides mechanisms to navigate through patches and view them in different formats, respond to user inputs, and run test simulations. *)
+
+val view : Patch.t list -> Nottui.ui Lwd.t
+(** [view patches] constructs a reactive UI for navigating and viewing a list of patches.
+    It supports toggling between help view and normal view, and handling navigation commands.
+    @param patches A list of patches to be displayed and navigated through.
+    @return A reactive UI element constructed using Lwd combinators. *)
+
```
panglesd, line 8: This function is internal to this module and is actually an implementation detail. It should not be part of the interface.

For instance, suppose we decide to change the internal implementation of diffcessible and stop using `Nottui`. In this case, this function would change type, making it an unnecessary breaking change.

lib/InteractiveViewer.mli
```diff
@@ -0,0 +1,21 @@
+(** This module defines the main user interface and interaction logic for a diff viewer application using the Nottui library. It provides mechanisms to navigate through patches and view them in different formats, respond to user inputs, and run test simulations. *)
```
panglesd, line 1: "Using the nottui library" is an implementation detail, irrelevant to the user here! ;)

lib/PatchNavigation.mli
```diff
@@ -0,0 +1,24 @@
...
+
+type direction =
+  | Prev
+  | Next
+      (** Type representing navigation directions within a list of patches. 
+          - [Prev] indicates navigation to the previous item.
+          - [Next] indicates navigation to the next item. *)
+
+val quit : bool Lwd.var
+(** [quit] is a variable indicating whether the application should terminate its main loop. *)
```
panglesd, line 11: Shouldn't this be in InteractiveViewer?

It seems to be unrelated to patch navigation to me!

lib/PatchNavigation.mli
```diff
@@ -0,0 +1,24 @@
...
+
+val navigate : 'a Zipper.t Lwd.var -> direction -> unit
+(** [navigate z_patches dir] navigates within a zipper of patches in the specified direction [dir].
+    It updates the zipper's state to the previous or next item based on the direction.
+    @param z_patches A reactive variable containing a zipper of patches.
+    @param dir The direction to navigate within the zipper.
+    @return () Updates the state of [z_patches] without returning any value. *)
+
+val change_summary : Patch.t Zipper.t Lwd.var -> Nottui.ui Lwd.t
```
panglesd, line 20: This does not seem to be related with patch navigation. Either it should be moved, or the name of this module should be more general!

Alsto, this should not take a `'a Zipper.t Lwd.t`: the change summary has no reason to modify the variable!
I think the best signature for this would be: `Patch.t Zipper.t -> Nottui.ui`

lib/OperationView.mli
```diff
@@ -0,0 +1,13 @@
+(** This module defines UI components for displaying information about operations in a diff viewer using the Nottui library. These components visualize different types of patch operations like creation, deletion, and edits. *)
+
+val operation_info : Patch.t Zipper.t Lwd.var -> Nottui.ui Lwd.t
```
panglesd, line 3: This should have type: `Patch.t Zipper.t -> Nottui.ui`.

lib/OperationView.mli
```diff
@@ -0,0 +1,13 @@
...
+(** [operation_info z_patches] creates a reactive UI element that displays information about the current operation in a zipper of patches.
+    It shows the current operation index, the total number of operations, and the number of hunks in the current patch.
+    @param z_patches A reactive variable containing a zipper of patches.
+    @return A reactive UI element that displays operation-related information. *)
+
+val current_operation : Patch.t Zipper.t Lwd.var -> Nottui.ui Lwd.t
```
panglesd, line 9: This should have type `Patch.t Zipper.t -> Nottui.ui`

lib/HunkView.mli
```diff
@@ -0,0 +1,23 @@
+(** This module provides functionalities for rendering different views of hunks within a patch in a diff viewer application using the Nottui library. It supports both unified and side-by-side display of changes. *)
+
+val current_hunks : Patch.t Zipper.t Lwd.var -> Nottui.ui Lwd.t
```
panglesd, line 3: Here again, the Lwd parts of this function should be removed!

lib/HunkView.mli
```diff
@@ -0,0 +1,23 @@
...
+(** [current_hunks z_patches] creates a reactive UI element to display all hunks in the current patch from a zipper of patches in a unified format.
+    @param z_patches A reactive variable containing a zipper of patches.
+    @return A reactive UI element listing all hunks in the current patch. *)
+
+val current_hunks_side_by_side : Patch.t Zipper.t Lwd.var -> Nottui.ui Lwd.t
```
panglesd, line 8: Here again, the Lwd parts of this function should be removed!

lib/HunkView.mli
```diff
@@ -0,0 +1,23 @@
...
+    This view aligns changes horizontally, allowing easier comparison of 'mine' and 'their' versions.
+    @param z_patches A reactive variable containing a zipper of patches.
+    @return A reactive UI element listing all hunks in the current patch in side-by-side view. *)
+
+type view_mode =
+  | SideBySide
+  | Normal
+      (** Defines available view modes for displaying diffs: either side-by-side or as a unified view. *)
+
+val view_mode : view_mode Lwd.var
```
panglesd, line 19: The logic for toggling view mode should either be all in this module, or all in another module, but not spread in multiple files!

lib/HunkView.ml
```diff
@@ -0,0 +1,178 @@
...
+            updated_their_acc)
+
+let lines_with_numbers lines attr_change prefix =
+  let line_num = ref 0 in
+  List.fold_left
+    (fun acc line ->
+      match line with
+      | Common s ->
+          incr line_num;
+          let content = Printf.sprintf "%3d   %s" !line_num s in
```
panglesd, line 128: Now all line numbers are considered to have 3 digits: that's a lot! :)

But no need to change anything in this PR: you have opened #66 anyway.

lib/HunkView.ml
```diff
@@ -0,0 +1,178 @@
...
+      | `Common line :: t ->
+          let common_mine_acc = Common line :: updated_mine_acc in
+          let common_their_acc = Common line :: updated_their_acc in
+          split_and_align_hunk t common_mine_acc common_their_acc
+      | _ ->
+          split_and_align_hunk remaining_hunks updated_mine_acc
+            updated_their_acc)
+
+let lines_with_numbers lines attr_change prefix =
+  let line_num = ref 0 in
```
panglesd, line 122: Let's not use a reference here!

panglesd (commented):
Looking good!

I made a series of comments. Refactoring in separate modules is not an easy task! We can go through them in a pair-programming session.

A minor comment that I cannot make inline is also that `lib/zipper.ml` (and maybe `bin/main.ml`) should follow the same naming convention as other files (so you should capitalize it).

lib/HelpView.mli
yokurang: Move help_visible variable into InteractiveViewer.ml

lib/HunkView.ml
yokurang: Use fold or something

lib/HunkView.mli
yokurang: We should define how to turn Patch.t Zipper -> Nottui.ui

Should not have Lwd stuff

The Lwd part should be extracted into InteractiveViewer.ml

lib/InteractiveViewer.mli
yokurang: Edit the comment

lib/InteractiveViewer.mli
yokurang: Just write intent of the function or important details that the function relies on to work 

If the function is obvious, we can only put a few words to keep the documentation minimal

lib/InteractiveViewer.ml
```diff
@@ -0,0 +1,120 @@
...
+                 | `ASCII 'q', [] ->
+                     toggle_help_visibility ();
+                     `Handled
+                 | _ -> `Unhandled)
+               (W.string "Type 'q' to exit the help panel");
+        ]
+    else
+      W.vbox
+        [
+          Lwd.pure (OperationView.operation_info (Lwd.peek z_patches_var));
```
yokurang, line 57: There is a bug where the changes are not reactive.. I think the solution lies somewhere in Lwd.map but after trying I can't seem to figure it out. Can I ask your advice on what I should do to resolve this? Thanks

lib/InteractiveViewer.ml
```diff
@@ -36,6 +36,10 @@ let view (patches : Patch.t list) =
       Lwd.set curr_scroll_state { state with position = state.W.bound }
     else Lwd.set curr_scroll_state state
   in
+  let project_op (op : Patch.t Zipper.t -> Nottui.ui) : ui Lwd.t =
+    let$ z_patches = Lwd.get z_patches_var in
+    op z_patches
```
Julow, line 41: I think you rewrote `Lwd.map`:
```
          Lwd.map (Lwd.get z_patches_var) ~f:OperationView.operation_info;
```

Julow: To clarify: `let$` is an alias to `Lwd.map`. This can be written in a shorter way without introducing a local function.

`Lwd.get` "observes" a variable, it returns a `Lwd.t` that changes every time the variable is changed.

lib/InteractiveViewer.ml
```diff
@@ -0,0 +1,126 @@
...
+  in
+  let curr_scroll_state = Lwd.var W.default_scroll_state in
+  let change_scroll_state _action state =
+    let off_screen = state.W.position > state.W.bound in
+    if off_screen then
+      Lwd.set curr_scroll_state { state with position = state.W.bound }
+    else Lwd.set curr_scroll_state state
+  in
+  let project_op (op : Patch.t Zipper.t -> Nottui.ui) : ui Lwd.t =
+    Lwd.map (Lwd.get z_patches_var) ~f:op
```
yokurang, line 40: @Julow I rewrote it so more explicit and easier to follow

Julow: I don't think this is easier to follow than inlining the `Lwd.map` directly where it's needed. I'd say introducing the `project_op` makes things less explicit.
We can keep it like that for now though.

lib/HelpView.mli
```diff
@@ -0,0 +1,5 @@
+val help_panel : Nottui.ui
+(** [help_panel] is the help panel, containing the help text. *)
+
+val help_keyboard_area : Nottui.ui
+(** [help_keyboard_area] is the help keyboard area, containing the keyboard shortcut to leave the help panel view. *)
```
panglesd, line 5: This is not used because the logic for it is in `InteractiveViewer.ml`

```suggestion
```

lib/HelpView.ml
```diff
@@ -0,0 +1,23 @@
...
+      W.string ~attr:shortcut_attr
+        "p:   Move to the previous operation, if present";
+      W.string ~attr:shortcut_attr "t:   Toggle view mode";
+      W.string ~attr:shortcut_attr "l:   Toggle line numbers";
+    ]
+
+let help_keyboard_area =
+  Ui.keyboard_area
+    (function `ASCII 'q', [] -> `Handled | _ -> `Unhandled)
+    (W.string ~attr:Notty.A.(fg lightblue) "Type 'q' to exit the help panel")
```
panglesd, line 23:
```suggestion
```

lib/HelpView.mli
```diff
@@ -0,0 +1,5 @@
+val help_panel : Nottui.ui
+(** [help_panel] is the help panel, containing the help text. *)
+
+val help_keyboard_area : Nottui.ui
```
Julow, line 4: `help_keyboard_area` is not used outside of this module.

lib/InteractiveViewer.ml
panglesd: I think it is because to:
- Call `Lwd.get` once to get an `Lwd.t` value that reacts on `z_patches_var` changes
- Inline this function on places where it is used.

Julow (approved):
This is a general improvement, let's merge !

lib/InteractiveViewer.ml
yokurang: Added

PR #67, opened 2024-07-01
added extunix specific ttyname for pager functionality

yokurang:


yokurang:
Try using `cat example.diff | dune exec ./_build/install/default/bin/diffcessible`

panglesd:
Running this without CLI options seems to break my terminal ^^
```
$ dune exec diffcessible
<empty non reactive terminal, killing diffcessible leaves the terminal broken>
```

yokurang:
When calling `./_build/install/default/bin/diffcessible` or if anything goes wrong, start the app with an empty diff content with title Error

bin/main.ml
```diff
@@ -1,18 +1,56 @@
...
+  let input_content =
+    match ic_opt with
+    | None -> create_empty_diff_content ()
+    | Some ic -> safe_read_input ic
+  in
+  let patch = Patch.to_diffs input_content in
+  Interactive_viewer.start ~term patch;
+  Notty_unix.Term.release term;
+  (* restore terminal to original state *)
+  Unix.tcsetattr Unix.stdin Unix.TCSANOW (Unix.tcgetattr Unix.stdin)
```
Julow, line 40: Why doesn't Notty do that ? Also, what would `Unix.tcgetattr` be when stdin is not a terminal ?

bin/main.ml
```diff
@@ -1,18 +1,56 @@
...
 let main file_path =
-  let file = In_channel.open_bin file_path in
-  let s = In_channel.input_all file in
-  let patch = Patch.to_diffs s in
-  Interactive_viewer.start patch
+  let ic_opt, term = setup_input file_path in
+  let input_content =
+    match ic_opt with
+    | None -> create_empty_diff_content ()
+    | Some ic -> safe_read_input ic
```
Julow, line 34: That might close stdin depending on what `setup_input` returns.

bin/main.ml
```diff
@@ -1,18 +1,56 @@
...
+  "diff --git a/Error b/Error\n--- a/Error\n+++ b/Error\n"
+
+let setup_input = function
+  | Some path ->
+      if not (Sys.file_exists path) then failwith ("File not found: " ^ path);
+      let ic = In_channel.open_bin path in
+      (Some ic, Notty_unix.Term.create ())
+  | None ->
+      let tty_path = ttyname Unix.stdout in
+      let tty_fd = Unix.openfile tty_path [ Unix.O_RDWR ] 0o600 in
```
Julow, line 15: `ttyy_fd` is never closed.

yokurang: Oops, forgot about that one

bin/main.ml
yokurang: Sorry can I clarify what you mean here? I thought `safe_read_input` handles closing stdin when it is opened?

Julow: `safe_read_input` always closes the channel but depending on which branch is taken in `setup_input`, the channel might be `stdin`, which shouldn't get closed.

bin/main.ml
Julow: To clarify, sometimes `stdin` might be a file or an other process stdout. It wouldn't make sense to change the terminal attribute of this. Also, as you query the attributes and set those again, I'm surprised this have any effect on the terminal state.

bin/main.ml
Julow: You just added a `close` after the `Notty_unix.Term.t` is created but before the program could run. That's not right, the file descriptor should be open until the program closes.

bin/main.ml
```diff
@@ -1,18 +1,55 @@
...
+  let ic_opt, term, tty_fd_opt = setup_input file_path in
+  let input_content =
+    match ic_opt with
+    | None -> create_empty_diff_content ()
+    | Some ic -> safe_read_input ic (ic == In_channel.stdin)
+  in
+  let patch = Patch.to_diffs input_content in
+  Interactive_viewer.start ~term patch;
+  Notty_unix.Term.release term;
+  match tty_fd_opt with Some tty_fd -> Unix.close tty_fd | None -> ()
```
yokurang, line 39: @Julow What do you think of this strategy to handle closing tty_fd?

bin/main.ml
```diff
@@ -1,18 +1,55 @@
...
+      let ic = In_channel.open_bin path in
+      (Some ic, Notty_unix.Term.create (), None)
+  | None ->
+      let tty_path = ttyname Unix.stdout in
+      let tty_fd = Unix.openfile tty_path [ Unix.O_RDWR ] 0o600 in
+      let term = Notty_unix.Term.create ~input:tty_fd ~output:tty_fd () in
+      if not (Unix.isatty Unix.stdin) then (Some In_channel.stdin, term, None)
+      else (None, term, Some tty_fd)
+
+let safe_read_input ic is_stdin =
```
yokurang, line 20: I believe this follows your comment on closing only if not stdin?

bin/main.ml
Julow: That's the right time to close the FD but it doesn't look very safe. The FD wouldn't be closed in case of an exception.

bin/main.ml
Julow: This seems to behave correctly but looks too complex to keep maintaining in my opinion. 

The more conventional pattern is the [`Fun.protect` function](https://ocaml.org/manual/5.2/api/Fun.html#VALprotect). It handles the exception in a more readable way, avoiding duplicating code. It can also combine well, for example, you could define a `with_input_fd` function with signature `file_path:string option -> (in_channel -> 'a) -> 'a`. This function would handle opening, calling the passed function and closing but nothing more.

The same pattern can be used for the `Term` value, can you also have a try?

yokurang: Use 

```
val with_open_bin : string -> ([t](https://ocaml.org/manual/5.2/api/In_channel.html#TYPEt) -> 'a) -> 'a
with_open_bin fn f opens a channel ic on file fn and returns f
    ic. After f returns, either with a value or by raising an exception, ic is guaranteed to be closed.
```

yokurang: One problem is reading from a closed file descriptor 

The goal is to eliminate fear of reading from a closed channel

yokurang:
TODO: Add cram tests for using `cat example.diff | dune exec ./_build/install/default/bin/diffcessible` and `dune exec ./_build/install/default/bin/diffcessible`

bin/main.ml
```diff
@@ -1,18 +1,58 @@
...
+  "diff --git a/Error b/Error\n--- a/Error\n+++ b/Error\n"
+
+let with_input_fd file_path f =
+  match file_path with
+  | Some path ->
+      if not (Sys.file_exists path) then failwith ("File not found: " ^ path);
+      In_channel.with_open_bin path f
+  | None -> f In_channel.stdin
+
+let setup_term () =
```
yokurang, line 15: Change setup_term to be with_setup_term using Fun.protect 

The main function does not need to be aware of tty_fd 

With the Fun.protect pattern to closed in the code that opens it

bin/main.ml
```diff
@@ -1,18 +1,58 @@
...
+  let tty_path = ttyname Unix.stdout in
+  let tty_fd = Unix.openfile tty_path [ Unix.O_RDWR ] 0o600 in
```
yokurang, line 17: Not needed if stdin is a tty 

Use if condition with Unix.isatty Unix.stdin

bin/main.ml
```diff
@@ -1,18 +1,58 @@
...
+  | Some path ->
+      if not (Sys.file_exists path) then failwith ("File not found: " ^ path);
+      In_channel.with_open_bin path f
+  | None -> f In_channel.stdin
+
+let with_setup_term f =
+  let tty_path = ttyname Unix.stdout in
+  let tty_fd = Unix.openfile tty_path [ Unix.O_RDWR ] 0o600 in
+  let term = Notty_unix.Term.create ~input:tty_fd ~output:tty_fd () in
+  let is_tty = Unix.isatty Unix.stdin in
```
Julow, line 19: Here you check whether `stdin` is a terminal. But if it is, you still open `tty_path`, that's not necessary!

bin/main.ml
```diff
@@ -1,18 +1,58 @@
...
+  with_setup_term (fun is_tty term ->
+      let input_content =
+        if is_tty then
+          match file_path with
+          | Some _ -> with_input_fd file_path read_input
+          | None -> create_empty_diff_content ()
+        else
+          Fun.protect
+            (fun () -> read_input In_channel.stdin)
+            ~finally:(fun () -> In_channel.close In_channel.stdin)
```
Julow, line 39: `stdin` do not need to be closed!

bin/main.ml
```diff
@@ -1,18 +1,58 @@
...
-  Interactive_viewer.start patch
+  with_setup_term (fun is_tty term ->
+      let input_content =
+        if is_tty then
+          match file_path with
+          | Some _ -> with_input_fd file_path read_input
+          | None -> create_empty_diff_content ()
+        else
+          Fun.protect
+            (fun () -> read_input In_channel.stdin)
```
Julow, line 38: This branch do not match on `file_path`, which might ignore the file passed as an argument.

bin/main.ml
yokurang: Right, sorry I confused myself a bit

bin/main.ml
```diff
@@ -1,18 +1,60 @@
 open Diffcessible
 open Cmdliner
+open ExtUnix.Specific
+
+let create_empty_diff_content () =
+  "diff --git a/Error b/Error\n--- a/Error\n+++ b/Error\n"
+
+let with_input_fd file_path f =
+  match file_path with
+  | Some path ->
```
Julow, line 10: No need to check whether `file_path` is `Some` or `None` here, this function is only ever called with `Some path`. It can be simplified.

bin/main.ml
```diff
@@ -1,18 +1,60 @@
...
+            else read_input In_channel.stdin
+      in
+      let patch = Patch.to_diffs input_content in
+      InteractiveViewer.start ~term patch)
 
 let file_arg =
-  let doc = "Path to the file containing the Git diff." in
-  Arg.(required & pos 0 (some string) None & info [] ~docv:"FILE" ~doc)
+  let doc = "Path to the input file." in
+  Arg.(value & opt (some string) None & info [ "f"; "file" ] ~docv:"FILE" ~doc)
```
Julow, line 51: Why changing to an `-` option ? A positional argument was feeling better I think.

Julow (approved):
I still have some comments but I'm very happy with the feature !

I would still like the change to an `-f` option instead of a positional argument to be removed. After that, feel free to merge!

lib/HunkView.ml
```diff
@@ -0,0 +1,184 @@
...
+type line = Change of string | Common of string | Empty
+
+(*
+
+    [Common, Their, Their, Mine]
+
+   List A [Common, Their, Their]
+   List B [Common, Mine, Empty]
+*)
+
```
panglesd, line 90:
```suggestion
```

panglesd (approved):
This looks ready to be merged!

Julow (approved):
Perfect, let's merge!

yokurang:
Squashed and merged

PR #72, opened 2024-07-09
Word Diff Algorithm Implementation Inspired by Myres Diff in Linear Space Theory

yokurang:
To test, run 
```
ocaml lib/DiffAlgorithms.ml
```

Will see
```
Foo
+ baz
- bar
+ bar
- baz
```


yokurang:
Some repetitive code but that is intentional as this is a draft still (WIP)

yokurang:
Current algorithm in mind for word-diff implementation: 

1. Obtain the list of diff changes. For reference, here is an example of the output from `get_diff` function from `simple-diff`.
```
let old_lines = [| "I"; "really"; "like"; "icecream" |]
let new_lines = [| "I"; "do"; "not"; "like"; "icecream" |]

module Diff = Simple_diff.Make(String)
open Diff

let expected =
  [ Equal   [| "I"; |];
    Deleted [| "really" |];
    Added [| "do"; "not" |];
    Equal [| "like"; "icecream" |];
]
```
2. Similar to the function `split_and_align_hunk` which returns two lists of "added" and "removed" diffs padded with the `Empty` type, implement a variation for word diffs in which the return type is two lists of lists of string (`string list list`). 

3. Implement a recursive function to process the `string list list` such that for each `string list` we apply the longest common subsequence function on both lists. All words that are part of the LCS will be tagged as `CommonWord`, and those that are not part of the LCS will be tagged as `AddedWord` or `DeletedWord` respectively. 

4. Convert this into Nottui elements.

PR #74, opened 2024-07-17
Word Diff Implementation for Normal View

yokurang:


yokurang:
Clean up later

yokurang:
Sample result:

yokurang:
Sample result:
image

lib/HunkView.ml
```diff
@@ -31,44 +110,54 @@ let ui_hunk_summary (hunk : Patch.hunk) : Nottui.ui =
       at_symbols;
     ]
 
-let rec process_lines (mine_num : int) (their_num : int) (acc : Nottui.ui list)
-    = function
+let process_line (hunk : Patch.hunk) i line =
+  let line_number =
+    W.string
+      ~attr:Notty.A.(fg yellow)
+      (Printf.sprintf "%4d " (succ (hunk.Patch.mine_start + i)))
```
Julow, line 117: This removes the "their" line number. I think that's not right.

lib/HunkView.ml
```diff
@@ -31,44 +110,54 @@ let ui_hunk_summary (hunk : Patch.hunk) : Nottui.ui =
...
+  in
+  match line with
+  | Common s ->
+      let content = W.string s in
+      Ui.hcat [ line_number; content ]
```
Julow, line 122: Previously, this function returned `(line_num_mine, line_num_their, ui)`, which was better as it could be shared with the split view.
Implementing word-diff in the split-view (in a later PR) will cause duplication due to this.

lib/HunkView.ml
```diff
@@ -31,44 +110,54 @@ let ui_hunk_summary (hunk : Patch.hunk) : Nottui.ui =
...
+            | WordDiff.WEqual words -> W.string (WordDiff.words_to_string words))
+          word_diffs
+      in
+      let space = W.string " " in
+      let content =
+        Ui.hcat
+          (List.fold_left (fun acc ui -> ui :: space :: acc) [] word_uis
+          |> List.rev)
+      in
+      Ui.hcat [ line_number; content ]
```
Julow, line 144: I find this rendering to be confusing. For a first version of the feature we definitely want to render the diffs as before, with removed and added lines. The word diff information should be used to highlight words within these lines.

lib/HunkView.ml
yokurang: This one is a big one:

Alan: Removed "their" number lines because the word-diff is now in one line
Jules: We want to keep the + and - lines and word-diff into these + and - lines the word highlighting 

Word-diff is something on top of regular diff view -> not something different

lib/HunkView.ml
```diff
@@ -1,7 +1,86 @@
...
+(* Common Functions *)
+
+type line =
+  | Change of string * WordDiff.word_diff list
+  | Common of string
+  | Empty
+
+let line_to_string = function Change (s, _) -> s | Common s -> s | Empty -> ""
+
+let construct_hunk mine their =
```
yokurang, line 13: This can be refactored into the word-diff module and needs some changes to process and construct the hunk properly as in the other comment.

lib/HunkView.ml
```diff
@@ -1,7 +1,86 @@
 open Nottui
 module W = Nottui_widgets
 
-(* Implementation of Single View Mode *)
+(* Common Functions *)
+
+type line =
+  | Change of string * WordDiff.word_diff list
```
yokurang, line 7: We need the assembled_line type to complete the implementation.

yokurang:
TODOs:

1. Fix up the type for line
2. Refactor and modify the rendering to match our specification 
3. Consequently match `construct_hunk` to match what we need

yokurang:
Example:
image

example.diff
```diff
@@ -1,3 +1,52 @@
+--- a/document.txt
```
Julow, line 1: Can you add the new examples at the bottom of the file ? This impacts the tests in a way that hides the impact of the whole PR.

lib/HunkView.ml
```diff
@@ -1,7 +1,86 @@
...
+        let s1 = line_to_string l1 in
+        let s2 = line_to_string l2 in
+        let word_diff = WordDiff.apply_word_diff s1 s2 in
+        aux (Change (s1, word_diff) :: acc) rest1 rest2
+    | [], their_lines ->
+        let mapped_lines =
+          List.map
+            (function
+              | Common s ->
+                  Change (s, [ WordDiff.WAdded (WordDiff.string_to_words s) ])
```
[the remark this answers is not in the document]
Julow: This comment is still valid. The function `create_line_diff` still creates `WordDiff` values, which should be moved inside the `WordDiff` module.

lib/HunkView.ml
```diff
@@ -31,43 +67,38 @@ let ui_hunk_summary (hunk : Patch.hunk) : Nottui.ui =
...
-      process_lines new_mine new_their (ui_element :: acc) rest
+let process_lines hunk =
+  let mine, their = split_and_align_hunk hunk.Patch.lines in
+  let create_line_diff m t =
+    match (m, t) with
+    | Common m_text, Common t_text when m_text = t_text ->
+        WordDiff.CommonDiff m_text
+    | Common m_text, Common t_text ->
+        let diff = WordDiff.Diff.apply_word_diff m_text t_text in
+        WordDiff.ModifiedDiff (diff, diff)
```
Julow, line 78: Isn't it weird that two aligned `Common` line are different ?
This cannot happen if `split_and_align_hunk` is correct and I think this impossible case could be eliminated if `split_and_align_hunk` was not used when computing word diffs.

Instead, collecting consecutive `Their` and `Mine` lines should be enough to compute word diffs. This would result in an API that is easier to understand and allows to reconstruct a value that looks a lot like the `hunk` type. My dream API for this would be:
```ocaml
val WordDiff.compute : Patch.hunk -> WordDiff.hunk
```
The `hunk` type is very easy to output, so a similar `WordDiff.hunk` type that contain the same information + the word diffs would be ideal to separate the code that computes the word diffs from the code that outputs them.

example.diff
```diff
@@ -1,3 +1,52 @@
...
++++ b/document.txt
+@@ -1,5 +1,5 @@
+-The quick brown fox jumps over the lazy dog.
++The fast brown fox leaps over the sleeping dog.
+ She sells seashells by the seashore.
+-The rain in Spain stays mainly in the plain.
++The rain in France falls mainly on the plain.
```
Julow, line 8: This example diff shows a bug:


I don't know what causes this but I think it's probably related to the lack of separation between the code that computes the word diffs from the code that output them.

lib/HunkView.ml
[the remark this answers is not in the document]
Julow: Let's close this thread as it's contained in the other comments I've made.

lib/WordDiff.ml
```diff
@@ -0,0 +1,147 @@
...
+  let words1 = Array.to_list (string_to_words s1) in
+  let words2 = Array.to_list (string_to_words s2) in
+  let common = lcs words1 words2 in
+  let rec construct_diff w1 w2 lcs acc =
+    match (w1, w2, lcs) with
+    | [], [], [] -> List.rev acc
+    | x :: xs, y :: ys, z :: zs when x = z && y = z ->
+        construct_diff xs ys zs (WEqual [| z |] :: acc)
+    | x :: xs, ys, lcs when not (List.mem x lcs) ->
+        construct_diff xs ys lcs (WDeleted [| x |] :: acc)
```
panglesd, line 50: Why do you need to check whether $x\in\texttt{lcs}$? This seems very costly, and wrong!

lib/WordDiff.ml
```diff
@@ -0,0 +1,147 @@
...
+    | xs, y :: ys, lcs when not (List.mem y lcs) ->
+        construct_diff xs ys lcs (WAdded [| y |] :: acc)
+    | [], [], zs -> construct_diff [] [] zs (WEqual (Array.of_list zs) :: acc)
+    | xs, [], [] -> construct_diff xs [] [] (WDeleted (Array.of_list xs) :: acc)
```
panglesd, line 54: Can this case really happen?

lib/WordDiff.ml
```diff
@@ -0,0 +1,147 @@
...
+  let rec construct_diff w1 w2 lcs acc =
+    match (w1, w2, lcs) with
+    | [], [], [] -> List.rev acc
+    | x :: xs, y :: ys, z :: zs when x = z && y = z ->
+        construct_diff xs ys zs (WEqual [| z |] :: acc)
+    | x :: xs, ys, lcs when not (List.mem x lcs) ->
+        construct_diff xs ys lcs (WDeleted [| x |] :: acc)
+    | xs, y :: ys, lcs when not (List.mem y lcs) ->
+        construct_diff xs ys lcs (WAdded [| y |] :: acc)
+    | [], [], zs -> construct_diff [] [] zs (WEqual (Array.of_list zs) :: acc)
```
panglesd, line 53: What values can `zs` take here?

lib/WordDiff.ml
```diff
@@ -44,16 +44,21 @@ let diff_words s1 s2 =
...
+        | true, false ->
+            construct_diff (x :: xs) ys (z :: zs) (WAdded [| y |] :: acc)
+        | false, false ->
+            construct_diff xs ys (z :: zs)
+              (WDeleted [| x |] :: WAdded [| y |] :: acc))
+    | x :: xs, [], lcs -> construct_diff xs [] lcs (WDeleted [| x |] :: acc)
+    | [], y :: ys, lcs -> construct_diff [] ys lcs (WAdded [| y |] :: acc)
+    | x :: xs, y :: ys, [] ->
+        construct_diff xs ys [] (WDeleted [| x |] :: WAdded [| y |] :: acc)
+    | [], [], z :: zs -> construct_diff [] [] zs (WEqual [| z |] :: acc)
```
panglesd, line 61: This case is not possible by construction! Instead of filling with some "never to be executed" code, you can write `assert false` which is the OCaml way of telling the reader "this branch cannot be taken, but OCaml is not able to see it."

yokurang: Ah, ok. I didn't really know what to put. Thanks !

panglesd (approved):
I left small comments, once they are resolved I think this is good to be merged!

PR #75, opened 2024-07-22
Towards word diffing

panglesd:
This PR adds a skeleton of functions and types that will be useful to implement word diffing.


panglesd (commented):
Nice !

Could you add a test? Also, you decide to represent the mine and their lines of a block in the reverse order, which is in my opinion counter-intuitive!

We might want to change the interface a little bit, to avoid the `mine_start`field.

PR #76, opened 2024-08-06
Towards rendering without colours

yokurang:


lib/Types.mli
```diff
@@ -0,0 +1,5 @@
+type rendering_mode = Color | TextMarkers
+type block_origin = Mine | Their | None
```
Julow, line 2: I'm not a fan of this module.

These types make more sense closer to where they are used and do not appear general enough to be used in several contextes. `block_origin` for example appears only in `Block`, which is now harder to read with this type in an other module. I think this applies to every types in this module.

An other example is that `rendering_mode` do not make sense inside `Block`, which depends on this module for other things.

lib/InteractiveViewer.ml
```diff
@@ -83,11 +92,15 @@ let view (patches : string Patch.t list) =
...
+                 | `ASCII 'r', [] ->
+                     toggle_rendering_mode ();
+                     `Handled
                  | _ -> `Unhandled)
                (W.string
                   "Type 'h' to go to the help panel, 'q' to quit, 'n' to go to \
                    the next operation, 'p' to go to the previous operation. \
-                   Press 't' to toggle view mode.");
+                   Press 't' to toggle view mode. Press 'r' to toggle \
+                   rendering mode.");
```
Julow, line 103: For a later PR! The help string is now wider than any reasonable terminal and is no longer readable. I think it should be rewritten in a more concise way, eg. `[h]elp [q]uit ...` or `[n/p] to navigate` or other variations. Also, it doesn't need to be exhaustive.

lib/HunkView.ml
```diff
@@ -124,14 +153,22 @@ let create_summary (start_line_num : int) (hunk_length : int)
...
 
-let ui_of_hunk_side_by_side (hunk : string Patch.hunk) : Nottui.ui =
-  let attr_mine = Notty.A.(fg red ++ st bold) in
-  let attr_their = Notty.A.(fg green ++ st bold) in
+let ui_of_hunk_side_by_side (hunk : string Patch.hunk)
+    (rendering_mode : Types.rendering_mode) : Nottui.ui =
+  let attr_mine, attr_their =
+    match rendering_mode with
+    | Types.Color -> (Notty.A.(fg red ++ st bold), Notty.A.(fg green ++ st bold))
+    | Types.TextMarkers -> (Notty.A.empty, Notty.A.empty)
```
Julow, line 161: `rendering_mode` is matched in too many places. For example, here it's matched to know what to pass to `lines_with_numbers`, which match it again for other reasons.
Also, it's matched several times in `ui_hunk_summary` to decides whether or not to apply attributes, which is also what this match does.

Julow (commented):
Nice :) The implementation is good but we need to iterate on the markers.

lib/WordDiff.ml
```diff
@@ -31,131 +29,189 @@ let diff_words (s1 : string) (s2 : string) : line_content * line_content =
...
 
-let word_to_ui word attr = W.string ~attr (word ^ " ")
+let word_to_ui word attr rendering_mode diff_type =
+  match rendering_mode with
+  | Types.Color -> W.string ~attr (word ^ " ")
+  | Types.TextMarkers -> (
+      match diff_type with
+      | `Equal -> W.string (word ^ " ")
+      | `Deleted -> W.string ("<-" ^ word ^ "/-> ")
+      | `Added -> W.string ("<+" ^ word ^ "/+> "))
```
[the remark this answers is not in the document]
alan-matthew-caladan: Refactor into a market variable.

lib/Types.mli
alan-matthew-caladan: Purpose: Avoid dependency cycles
But try something else.

lib/WordDiff.ml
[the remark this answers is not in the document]
panglesd: In another PR it might be good to consider "grouping" consecutive added/removed words using single annotations, especially when using "opening-and-closing" markers.

lib/HunkView.ml
```diff
@@ -82,76 +209,50 @@ let ui_unified_diff (hunk : string Patch.hunk) : Nottui.ui =
...
-    ]
-
-let current_hunks_side_by_side (z_patches : string Patch.t Zipper.t) : Nottui.ui
-    =
+let current_hunks_side_by_side (z_patches : string Patch.t Zipper.t)
+    (mode : rendering_mode) : Ui.t =
   let p = Zipper.get_focus z_patches in
-  let hunks_ui = List.map ui_of_hunk_side_by_side p.Patch.hunks in
-  Ui.vcat hunks_ui
+  let render_side_by_side hunk =
```
Julow, line 226: Common lines are now colored in the side-by-side view, which breaks it.

Julow (commented):
It would be nice to have the new output appearing in the tests.

test/cram/test.t
```diff
@@ -21,28 +21,28 @@ This is a cram test for the new executable.
...
-      5 +   let patch = Patch.to_diffs s in
-      6 +   Interactive_viewer.start patch
-   5  7   
-   6  8   open Cmdliner
-   7  9   
+   1  1   open Diffcessible 
+   2  2    
+   3  3   let main () = 
+   4    -   Interactive_viewer.start () 
+      4 +   let s = In_channel.input_all In_channel.stdin in   let patch = Patch.to_diffs s in   Interactive_viewer.start patch
```
Julow, line 34: This is not right, two lines are merged together. On main we have a check that avoids doing word diff in this case but it's removed in this PR.

lib/HunkView.ml
```diff
@@ -1,9 +1,15 @@
 open Nottui
 module W = Nottui_widgets
 
+(* Constants and Formatting Functions *)
+let added_marker (content : Ui.t) : Ui.t = Ui.hcat [ W.string "+"; content ]
+let removed_marker (content : Ui.t) : Ui.t = Ui.hcat [ W.string "-"; content ]
+let unchanged_marker (content : Ui.t) : Ui.t = Ui.hcat [ W.string ""; content ]
```
Julow, line 7: Is this any different than just returning `content` ?

lib/HunkView.ml
```diff
@@ -67,35 +91,91 @@ let ui_hunk_summary (hunk : string Patch.hunk) : Nottui.ui =
...
+    (diff_type : [ `Added | `Removed | `Unchanged ]) (mode : rendering_mode) :
+    Ui.t =
+  match mode with
+  | TextMarkers -> (
+      match diff_type with
+      | `Added -> W.string (Printf.sprintf "   %2d + " (their_num + 1))
+      | `Removed -> W.string (Printf.sprintf "%2d    - " (mine_num + 1))
+      | `Unchanged ->
+          W.string (Printf.sprintf "%2d %2d   " (mine_num + 1) (their_num + 1)))
+  | Color -> (
```
Julow, line 104: The only difference between these two match cases is that in one case `attr` is not passed ? These two cases should be factored to avoid a maintenance burden.

Here's a trick: It's possible to pass an `option` value as an optional argument:
```
let attr_added = None

W.string ?attr:attr_added "..."
```

lib/HunkView.ml
```diff
@@ -105,53 +185,90 @@ let lines_with_numbers (lines : line list) (attr_change : Notty.attr)
...
+let current_hunks_side_by_side (z_patches : string Patch.t Zipper.t)
+    (mode : rendering_mode) : Ui.t =
+  let p = Zipper.get_focus z_patches in
+  let render_side_by_side hunk =
+    let mine_lines, their_lines = split_and_align_hunk hunk.Patch.lines in
+    let render_lines lines marker attr =
+      lines_with_numbers lines attr marker mode
+    in
+    let content_mine =
+      render_lines mine_lines removed_marker Notty.A.(fg red)
```
Julow, line 248: `removed_marker` and `added_marker` are working with word in the context of word diff. It doesn't make sense to apply them to entire lines.

lib/HunkView.ml
Julow: That's fixed.

test/cram/test.t
```diff
@@ -21,28 +21,28 @@ This is a cram test for the new executable.
```
Julow, line 21: We should have a test for colorless in the side-by-side view. It's currently not right.

test/cram/test.t
yokurang: Resolved. Reverted cram tests to follow main

test/cram/test.t
yokurang: Not yet.

yokurang: Added additional tests

test/cram/test.t
Julow: Reverting the cram tests don't make sense, you need to run and promote them. If I run them, this bug appear again, so something is broken in the code.

lib/HunkView.ml
Julow: Why did you remove all the functions ?  These seemed useful.

lib/HunkView.ml
```diff
@@ -67,91 +89,197 @@ let ui_hunk_summary (hunk : string Patch.hunk) : Nottui.ui =
...
     | line :: rest ->
-        let content, attr, next_num =
+        let line_ui, next_num =
           match line with
           | Common s ->
               let content = Printf.sprintf "%3d   %s" line_num s in
-              (content, Notty.A.empty, line_num + 1)
+              let attr =
+                match mode with
+                | Color -> Some Notty.A.(fg lightcyan)
```
Julow, line 184: Turning common lines or words into light cyan is not related to the goal of this PR and should be reverted.

lib/HunkView.ml
yokurang: Hmm, I thoght that was what you meant by the comment..?

lib/HunkView.ml
```diff
@@ -67,91 +89,197 @@ let ui_hunk_summary (hunk : string Patch.hunk) : Nottui.ui =
...
-    in
-    if single_line_changes then
-      let word_diff_blocks = List.map WordDiff.compute blocks in
-      let word_diff_lines = Block.to_hunk word_diff_blocks in
-      WordDiff.render_hunk_lines word_diff_lines
-    else WordDiff.render_hunk hunk
+let render_line_number (mine_num : int) (their_num : int)
+    (diff_type : [ `Added | `Removed | `Unchanged ]) (mode : rendering_mode) :
+    Ui.t =
+  let attr_option =
```
Julow, line 95: An other optional argument tips: If you name this `attr`, you could pass it as `?attr` instead of `?attr:attr`.

I think this is applicable here as the longer name is not useful.

test/cram/test.t
yokurang: Got it

test/cram/test.t
```diff
@@ -379,7 +379,15 @@ This is a cram test for the new executable.
   
   Type 'h' to go to the help panel, 'q' to quit, 'n' to go to the next operation, 'p' to go to the previous operation. Press 't' to toggle view mode.
 
+# Test unified view with text markers (colorless)
+$ dummy_terminal example.diff r
```
Julow, line 383: The lines starting in `$` must be indented by two space if you want them to be executed.

test/cram/test.t
```diff
@@ -379,7 +379,15 @@ This is a cram test for the new executable.
...
 
+# Test side-by-side view with text markers (colorless)
+$ dummy_terminal example.diff t r
 
+# Test toggling between side-by-side and unified views
+$ dummy_terminal example.diff t t
 
+# Test toggling render mode (color/colorless)
+$ dummy_terminal example.diff r r
```
Julow, line 392: I'm not sure these two last tests are interesting. Instead, I'd suggest showing the hunks 3 and 4 in normal view with color less. `n n r` and `n n n r`.

lib/HunkView.ml
```diff
@@ -2,10 +2,14 @@ open Nottui
 module W = Nottui_widgets
 
 (* Types *)
-
 type line = Change of string | Common of string | Empty
+type rendering_mode = Color | TextMarkers
 
 (* Utility Functions *)
+let line_marker (sign : string option) (content : Ui.t) : Ui.t =
+  match sign with
```
Julow, line 10: This function is used only in one place, where `sign` is never `None`. Did you mean to also use it in other places ?

lib/HunkView.ml
Julow: Sorry for the confusion, my comment was about `unchanged_marker` only.

lib/HunkView.ml
yokurang: it was previously used in another place but I reverted that code in between my commits, so it is only used in one place now. Need to fix that

test/cram/test.t
Julow: To be clear, on `main` this renders as:

```
 1  1   open Diffcessible
 2  2   
 3  3   let main () =
 4    -   Interactive_viewer.start ()
    4 +   let s = In_channel.input_all In_channel.stdin in
    5 +   let patch = Patch.to_diffs s in
    6 +   Interactive_viewer.start patch
 5  7   
 6  8   open Cmdliner
 7  9   
```

With this PR, this now renders like this:

```
 1  1   open Diffcessible 
 2  2    
 3  3   let main () = 
 4    -   Interactive_viewer.start () 
    4 +   let s = In_channel.input_all In_channel.stdin in   let patch = Patch.to_diffs s in   Interactive_viewer.start patch 
 5  5    
 6  6   open Cmdliner 
 7  7    
```

This seems to be related to `render_hunk`. In `ui_unified_diff`, there's a check whether or not to enable word diffing. In the false case, `render_hunk` gets called, which compute word diffs anyway.
That might be a mistake during a refactoring. By the way, `render_hunk` has a clear purpose and might have a more descriptive name.

yokurang: Fixed

lib/HunkView.ml
```diff
@@ -37,25 +40,43 @@ let split_and_align_hunk hunks : line list * line list =
...
+  let styled_word =
+    match (mode, change_type) with
+    | Color, `Added -> W.string ~attr:Notty.A.(fg green) word
+    | Color, `Removed -> W.string ~attr:Notty.A.(fg red) word
+    | Color, `Unchanged -> W.string word
+    | TextMarkers, `Added -> added_marker (W.string word)
+    | TextMarkers, `Removed -> removed_marker (W.string word)
+    | TextMarkers, `Unchanged -> W.string word
+  in
+  Ui.hcat [ styled_word; W.string " " ]
```
panglesd, line 61: This is adding an extra space at the end of the line. However, this small bug is not introduced in this PR and should be fixed in another one.

panglesd (approved):
Looks good to me!

lib/HunkView.ml
yokurang: Ok

Julow (approved):
Looks good to me too :)

PR #80, opened 2024-08-20
Using Edit Distance to Word-Diff Across Multiple Lines

yokurang:


lib/WordDiff.ml
```diff
@@ -43,98 +77,67 @@ let edit_distance (type a) (compare : a -> a -> bool) (s : a array)
...
+            construct_pairs xs ys (z :: zs) ((Some x, Some y) :: acc))
     | x :: xs, [], lcs ->
-        construct_diff xs [] lcs (Changed x :: acc_mine) acc_their
+        construct_pairs xs [] lcs ((Some x, None) :: acc)
     | [], y :: ys, lcs ->
-        construct_diff [] ys lcs acc_mine (Changed y :: acc_their)
+        construct_pairs [] ys lcs ((None, Some y) :: acc)
     | x :: xs, y :: ys, [] ->
-        construct_diff xs ys [] (Changed x :: acc_mine) (Changed y :: acc_their)
+        construct_pairs xs ys [] ((Some x, Some y) :: acc)
```
panglesd: Pairing these lines is wrong!

yokurang: Hi, do you mind elaborating more on why it is wrong?

panglesd: Try it on the following test!
```diff
- aaaaaaaaa a
- egdg xsd fi sdgfyu s gd csue c e yeg scy
+ aaaaaaaaa b
+ yugyu fft yu utf ytfutf uyg gygu yftuf gg uf ft
```

yokurang: LACS -> A sequence of lines 
Here, we have two lines in 1 and two lines in 2. 
So LCAS is a matter of lines. Here, it will be a length of 1, and the LCAS are (Some aaaaaaaaa a, Some aaaaaaaaa b)

So here, in the accumulator, I am adding x and y as being paired, when x and y are not part of the LACS. 
Here, the example, x and y corresponds to egdg ... and yugyu ...
So they should not be paired.

